PR #83, opened 2023-05-09
Fix: Update annotation structure, put CAIS annotations under annotation["assetInfo"]

### LINKED ISSUE #72: Presubmit check should validate mapping annotations keys
https://github.com/abcxyz/pmap/blob/main/pkg/mapping/processors/assetinventory.go#L162-L182

We use these keys (from CAIS) in the annotations. If somebody manually provides an annotation that has the same key, it could get overwritten by the CAIS processor.
We need to validate that the mapping is not using such well-known annotation keys.
Even better, CAIS info probably should be fully under a dedicated annotation. E.g.

```
annotations:
  assetInfo:
    iamPolicies:
    labels: foo
```

cc: @capri-xiyue

sailorlqh:
fix #72

### DIFF
--- a/apis/v1alpha1/validation.go
+++ b/apis/v1alpha1/validation.go
@@ -31,5 +31,8 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
 	if m.Resource.Provider == "" {
 		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
 	}
+	if _, ok := m.Annotations.AsMap()["assetInfo"]; ok {
+		vErr = errors.Join(vErr, fmt.Errorf("input includes reserved key: assetInfo"))
+	}
 	return
 }
--- a/apis/v1alpha1/validation.go
+++ b/apis/v1alpha1/validation.go
@@ -32,7 +32,7 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
 		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
 	}
 	if _, ok := m.Annotations.AsMap()["assetInfo"]; ok {
-		vErr = errors.Join(vErr, fmt.Errorf("input includes reserved key: assetInfo"))
+		vErr = errors.Join(vErr, fmt.Errorf("reserved key is included: assertInfo"))
 	}
 	return
 }
--- a/apis/v1alpha1/validation.go
+++ b/apis/v1alpha1/validation.go
@@ -21,6 +21,11 @@ import (
 	"net/mail"
 )
 
+const (
+	// Reserved key where annotation from CAIS will be stored.
+	AnnotationKeyAssetInfo = "assetInfo"
+)
+
 // ValidateResourceMapping checks if the ResourceMapping is valid.
 func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
 	for _, e := range m.Contacts.Email {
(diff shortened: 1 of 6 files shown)


pkg/mapping/processors/assetinventory_test.go
```diff
@@ -143,33 +143,45 @@ func TestProcessor_UpdatedProcess(t *testing.T) {
 					Provider: "gcp",
 					Name:     "//pubsub.googleapis.com/projects/test-project/topics/test-topic",
 				},
-				Contacts: &v1alpha1.Contacts{Email: []string{"pmap.gmail.com"}},
+				Contacts: &v1alpha1.Contacts{Email: []string{"pmap@gmail.com"}},
```
yolocs, line 146: Can we globally change to `@example.com`

apis/v1alpha1/validation.go
```diff
@@ -31,5 +31,8 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
 	if m.Resource.Provider == "" {
 		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
 	}
+	if _, ok := m.Annotations.AsMap()["assetInfo"]; ok {
```
yolocs, line 34: Probably in this package - Can we define the well-known keys as (exposed) consts?

yolocs (changes requested):

apis/v1alpha1/validation.go
```diff
@@ -21,6 +21,8 @@ import (
 	"net/mail"
 )
 
+const WellKnownKey = "assetInfo"
```
capri-xiyue, line 24: nits: using 
```
const (
	WellKnownKey = "assetInfo"
)
```

apis/v1alpha1/validation_test.go
```diff
@@ -38,7 +38,7 @@ func TestValidateResouceMapping(t *testing.T) {
 					Name:     "//pubsub.googleapis.com/projects/test-project/topics/test-topic",
 				},
 				Contacts: &Contacts{
-					Email: []string{"pmap.gmail.com"},
+					Email: []string{"pmap.example.com"},
```
capri-xiyue, line 41: change to @

sailorlqh: I think this is the invalid email test, we want an invalid email here?

capri-xiyue: Sorry, I missed that(the previous code lines were hidden by default). Maybe change it to `invalid.example.com` to make it more obvious

sailorlqh: Done.

apis/v1alpha1/validation.go
capri-xiyue: Did you change the variable naming?

capri-xiyue (approved):

apis/v1alpha1/validation.go
```diff
@@ -21,6 +21,10 @@ import (
 	"net/mail"
 )
 
+const (
+	AssetInfoKey = "assetInfo"
```
yolocs, line 25: Go doesn't have enum so the convention is to put "type" as the first part of the variable name, e.g. `AnnotationKeyAssetInfo`, `AnnotationKeyOtherStuff`

yolocs (changes requested):

apis/v1alpha1/validation.go
sailorlqh: Thanks for these comments, I will keep them in-mind! Already updated.

apis/v1alpha1/validation.go
```diff
@@ -21,6 +21,10 @@ import (
 	"net/mail"
 )
 
+const (
+	AnnotationKeyAssetInfo = "assetInfo"
```
yolocs, line 25: Comment

yolocs (approved):

PR #85, opened 2023-05-10
feat: added test data

capri-xiyue:
added test data and workflow yaml needed for manual e2e test


.github/workflows/fake-bets-mapping-snapshot-file-change.yml
```diff
@@ -0,0 +1,16 @@
...
+concurrency: '${{ github.workflow }}-${{ github.head_ref || github.ref }}-fake-bets-mapping-snapshot-file-change'
+
+jobs:
+  snapshot:
+    uses: 'abcxyz/pmap/.github/workflows/snapshot-file-change.yml@main' # ratchet:exclude
+    with:
+      WIF_PROVIDER: 'projects/921163060412/locations/global/workloadIdentityPools/github-pool-3b39/providers/github-provider'
+      WIF_SERVICE_ACCOUNT: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
+      destination_prefix: 'pmap-dev/mapping'
+      path: 'test/e2e/testdata/fakebets/mapping'
```
sqin2019, line 16: Just curious, after this change is merged, how the changed file action will work in order to get the already added mapping files?

capri-xiyue: If there is no change, the changed file action won't copy the already added mapping files to gcs. You need to rely on `file-copy` one. If you want to see whether the file change still works, you'll have to change the mapping files, the easier change could be adding annotations.

sqin2019: Got it, thx!

sqin2019 (approved):

PR #88, opened 2023-05-11
fix: skip uploading when there is no changed yaml files

capri-xiyue:
skip uploading when there is no changed yaml files


yolocs (approved):
Could you put "skip uploading when there is no changed yaml files" in the PR title?

PR #89, opened 2023-05-11
fix: use a stoppable interface for processor

sqin2019:
Since not every processor is stoppable.

### DIFF
--- a/pkg/server/event_handler.go
+++ b/pkg/server/event_handler.go
@@ -52,6 +52,10 @@ type ProtoWrapper[T any] interface {
 // Processor[*structpb.Struct].
 type Processor[P proto.Message] interface {
 	Process(context.Context, P) error
+}
+
+// StoppableProcessor is the interface to processors that are stoppable.
+type StoppableProcessor[P proto.Message] interface {
 	Stop() error
 }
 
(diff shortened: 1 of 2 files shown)


capri-xiyue:
Should we ask every processor to implement stoppable method. If there is nothing to stop, just return nil?

sqin2019:
> Should we ask every processor to implement stoppable method. If there is nothing to stop, just return nil?

This is to avoid that, so that if nothing to stop, then no need to implement it.

capri-xiyue:
> > Should we ask every processor to implement stoppable method. If there is nothing to stop, just return nil?
> 
> This is to avoid that, so that if nothing to stop, then no need to implement it.

Personally, it is better to force users to implement a stop method in case they forget it, and return nil is kind of trivial. That's my two cents. Let's see what other people thinks about it.

yolocs (approved):

yolocs:
I'd like to see the "stop" logic handled outside of the handler. We pass in processors anyway. They can be stopped outside as well.

See: https://github.com/abcxyz/access-on-demand/pull/7#discussion_r1189215572 We ended up remove the "stop" logic from the handler.

sqin2019:
> I'd like to see the "stop" logic handled outside of the handler. We pass in processors anyway. They can be stopped outside as well.
> 
> See: [abcxyz/access-on-demand#7 (comment)](https://github.com/abcxyz/access-on-demand/pull/7#discussion_r1189215572) We ended up remove the "stop" logic from the handler.

Created a follow up.

Issue #21, opened 2023-02-09
Add additional metadata to pmap event.


sqin2019:
https://cloud.google.com/storage/docs/pubsub-notifications#payload
https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata

PR #93, opened 2023-05-12
Fix: add github metadata to PmapEvent

sailorlqh:
fix: #21


pkg/server/event_handler.go
```diff
@@ -293,6 +299,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
 	return p, nil
 }
 
+func (h *EventHandler[T, P]) getGCSObjectGithubResourceMetadata(ctx context.Context, objAttrs map[string]string) (*v1alpha1.GitHubSource, error) {
```
yolocs, line 302: Not need to be a struct func
```suggestion
func getGCSObjectGithubResourceMetadata(ctx context.Context, objAttrs map[string]string) (*v1alpha1.GitHubSource, error) {
```

yolocs, line 302: Prefer simpler names `parseGitHubSource(ctx, objAttrs)`

pkg/server/event_handler.go
```diff
@@ -293,6 +299,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	objectID, found := objAttrs["objectId"]
+	if !found {
+		return nil, fmt.Errorf("object ID not found")
+	}
+
+	// Read the object from bucket.
+	attrs, err := h.client.Bucket(bucketID).Object(objectID).Attrs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("object attrs not found")
+	}
```
yolocs, line 316: I don't think we need to call GCS again. The PubSub notification we got should already have the metadata. See: https://cloud.google.com/storage/docs/pubsub-notifications#payload

pkg/server/event_handler.go
```diff
@@ -293,6 +299,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	m := attrs.Metadata
+	r := &v1alpha1.GitHubSource{}
+	c, found := m["git-commit"]
```
yolocs, line 319: Can you help fix the keys here to make them all prefixed with "github", e.g. `github-commit`. git != github

pkg/server/event_handler.go
```diff
@@ -293,6 +299,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	}
+	w, found := m["git-workflow"]
+	if found {
+		r.Workflow = w
+	}
+	ws, found := m["git-workflow-sha"]
+	if found {
+		r.WorkflowSha = ws
+	}
+	t, found := m["triggered-timestamp"]
```
yolocs, line 335: Also prefix with "github-workflow-timestamp"

protos/v1alpha1/pmap_event.proto
```diff
@@ -44,4 +44,13 @@ message GitHubSource {
...
   string commit = 3;
+
+  // Required. The github workflow that triggered the pmap event.
+  string workflow = 4;
+
+  // Required. The sha for the github workflow.
+  string workflow_sha = 5;
+
+  // Required. The timestamp when workflow is triggered.
+  string triggered_timestamp = 6;
```
yolocs, line 55: This should be a [time?](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp)

`workflow_timestamp`

protos/v1alpha1/pmap_event.proto
```diff
@@ -44,4 +44,13 @@ message GitHubSource {
 
   // Required. The git commit.
   string commit = 3;
+
+  // Required. The github workflow that triggered the pmap event.
+  string workflow = 4;
+
+  // Required. The sha for the github workflow.
+  string workflow_sha = 5;
```
yolocs, line 52: cc: @capri-xiyue 

Should we add workflow run id and run attempt?

pkg/server/event_handler.go
```diff
@@ -293,6 +299,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	if err != nil {
+		return nil, fmt.Errorf("object attrs not found")
+	}
+	m := attrs.Metadata
+	r := &v1alpha1.GitHubSource{}
+	c, found := m["git-commit"]
+	if found {
+		r.Commit = c
+	}
+	rn, found := m["git-repo"]
```
yolocs, line 323: Where do we set `file_path`?

yolocs (changes requested):

apis/v1alpha1/pmap_event.pb.go
```diff
@@ -118,11 +118,17 @@ type GitHubSource struct {
 	unknownFields protoimpl.UnknownFields
 
 	// Required. The repository name where the payload is located.
-	RepoName string `protobuf:"bytes,1,opt,name=repo_name,json=repoName,proto3" json:"repo_name,omitempty"`
+	RepoName string `protobuf:"bytes,1,opt,name=repo_name,json=repoName,proto3" json:"repo-name,omitempty"`
```
capri-xiyue, line 121: https://pkg.go.dev/encoding/json  should use `json:"repoName,omitempty"` to follow the naming convention of json tag

apis/v1alpha1/pmap_event.pb.go
```diff
@@ -118,11 +118,17 @@ type GitHubSource struct {
...
 	// Required. The file path of the payload.
-	FilePath string `protobuf:"bytes,2,opt,name=file_path,json=filePath,proto3" json:"file_path,omitempty"`
+	FilePath string `protobuf:"bytes,2,opt,name=file_path,json=filePath,proto3" json:"file-path,omitempty"`
```
capri-xiyue, line 123: same

apis/v1alpha1/pmap_event.pb.go
```diff
@@ -118,11 +118,17 @@ type GitHubSource struct {
...
 	// Required. The git commit.
 	Commit string `protobuf:"bytes,3,opt,name=commit,proto3" json:"commit,omitempty"`
+	// Required. The github workflow that triggered the pmap event.
+	Workflow string `protobuf:"bytes,4,opt,name=workflow,proto3" json:"workflow,omitempty"`
+	// Required. The sha for the github workflow.
+	WorkflowSha string `protobuf:"bytes,5,opt,name=workflow_sha,json=workflowSha,proto3" json:"workflow-sha,omitempty"`
```
capri-xiyue, line 129: same

pkg/server/event_handler.go
sailorlqh: Xiyue and I talked about this last week. Currently, in the `file-change` and `file-copy workflows`, `file_path` is not uploaded to metadata. https://github.com/abcxyz/pmap/blob/main/.github/workflows/snapshot-file-change.yml#L73-L78
Our idea is for current PR, we only add this metadata, and there will be a separate PR that updates the workflow files to include `file_path`.
By the way, what do you think if we use that PR to fix the renaming and add more fields?

pkg/server/event_handler_test.go
```diff
@@ -92,6 +92,7 @@ func TestEventHandler_HttpHandler(t *testing.T) {
 		name               string
 		pubsubMessageBytes []byte
 		gcsObjectBytes     []byte
+		githubObjectBytes  []byte
```
capri-xiyue, line 95: The naming is confusing. Maybe gcsObjectMetadataBytes.

pkg/server/event_handler_test.go
```diff
@@ -325,18 +348,40 @@ func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.
...
+		  "git-workflow-sha": "test-workflow-sha",
+		  "git-workflow": "test-workflow",
+		  "git-repo": "test-git-repo"
+		}
+	  }`)
+}
+
 // Returns a fake http func that writes the data in http response.
-func handleObjectRead(t *testing.T, data []byte) func(w http.ResponseWriter, r *http.Request) {
+func handleObjectRead(t *testing.T, data, metadata []byte) func(w http.ResponseWriter, r *http.Request) {
```
capri-xiyue, line 365: I personally think it would be better if we can write a fake server like https://github.com/abcxyz/pmap/blob/41c46ce082d64d08396dd60bb3b5a48762f08bdd/pkg/mapping/processors/assetinventory_test.go#L79 for UT. The way to write unit test is a little bit difficulty to understand and implement as we need to figure out the url of gcs function call. And the switch statement of different url is a little bit cumbersome. But no a block for this PR. We can create another issue to improve this.

protos/v1alpha1/pmap_event.proto
capri-xiyue: Yes. That would be useful.

test/integration/main_test.go
```diff
@@ -129,6 +130,14 @@ func TestMappingEventHandling(t *testing.T) {
 	for _, tc := range cases {
 		tc := tc
 
+		wantGithubResource := &v1alpha1.GitHubSource{
+			RepoName:           "test-git-repo",
```
capri-xiyue, line 134: use github instead of git

test/integration/main_test.go
```diff
@@ -349,7 +378,7 @@ func testUploadFile(ctx context.Context, tb testing.TB, bucket, object string, d
 	o := gcsClient.Bucket(bucket).Object(object)
 
 	// For an object that does not yet exist, set the DoesNotExist precondition.
-	o = o.If(storage.Conditions{DoesNotExist: true})
+	// o = o.If(storage.Conditions{DoesNotExist: true})
```
capri-xiyue, line 381: Why do you comment this?

test/integration/main_test.go
```diff
@@ -177,6 +186,12 @@ contacts:
 				t.Errorf("resourcemapping(ignore annotation) unexpected diff (-want,+got):\n%s", diff)
 			}
 
+			gotGithubResouce := gotPmapEvent.GetGithubSource()
+
+			if diff := cmp.Diff(gotGithubResouce, wantGithubResource, cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{})); diff != "" {
```
capri-xiyue, line 191: write the gotPmapEvent.GetGithubSource() inline,

test/integration/main_test.go
```diff
@@ -262,6 +285,12 @@ deletion_timeline:
 			if diff := cmp.Diff(wantPayload, gotPayload, protocmp.Transform()); diff != "" {
 				t.Errorf("gotPayload unexpected diff (-want,+got):\n%s", diff)
 			}
+
+			gotGithubResouce := gotPmapEvent.GetGithubSource()
+
+			if diff := cmp.Diff(wantGithubResource, gotGithubResouce, cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{})); diff != "" {
```
capri-xiyue, line 291: same

pkg/server/event_handler.go
yolocs: Sure. It'd probably be easier if you do the other PR first to address the renaming and add file path

apis/v1alpha1/pmap_event.pb.go
yolocs: Good catch. Aren't these generated?

pkg/server/event_handler.go
capri-xiyue: To include `filepath` requires more workload, I'd prefer we handle it in another PR. We will need to change some logic in https://github.com/abcxyz/pmap/blob/main/.github/workflows/snapshot-file-copy.yml. Currently https://github.com/google-github-actions/upload-cloud-storage does not keep the all parent path of file when it uploads file to the gcs. https://github.com/abcxyz/pmap/blob/main/.github/workflows/snapshot-file-copy.yml need to follow some logic in https://github.com/abcxyz/pmap/blob/41c46ce082d64d08396dd60bb3b5a48762f08bdd/.github/workflows/snapshot-file-change.yml#L46 to record the info of `file_path` when it uploads file to gcs

yolocs: sgtm. Maybe the rename could be piggybacked here?

test/integration/main_test.go
sailorlqh: https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1724-L1727
Because the upload container to steps, first is write the object, second is add metadata, if DoseNotExist is set to true, the add metadata will fail because the this object is already there.

If we can to keep the conditions, then we would need a new writer to add metadata, I think that's not necessary.

pkg/server/event_handler_test.go
sailorlqh: Created issue #95 to solve this.

pkg/server/event_handler.go
sailorlqh: I will do the renaming, adding field `workflow run id`, `run attempt`, and include `file_path` in the next PR.

pkg/server/event_handler.go
sailorlqh: Will do in next PR, this is included in #94

test/integration/main_test.go
sailorlqh: Found a new way to write metadata which allows us to keep using preconditions.

pkg/server/event_handler_test.go
```diff
@@ -89,50 +90,58 @@ func TestEventHandler_HttpHandler(t *testing.T) {
 	ctx := context.Background()
 
 	cases := []struct {
-		name               string
-		pubsubMessageBytes []byte
-		gcsObjectBytes     []byte
-		wantStatusCode     int
-		wantRespBodySubstr string
+		name                   string
+		pubsubMessage          *PubSubMessage
```
capri-xiyue, line 94: It looks confusing to me, why do you need both pubsubMessage and PubsubMessageBytes?

pkg/server/event_handler_test.go
```diff
@@ -176,18 +193,20 @@ func TestEventHandler_Handle(t *testing.T) {
...
-		notification     pubsub.Message
-		gcsObjectBytes   []byte
-		processors       []Processor[*structpb.Struct]
-		successMessenger Messenger
-		failureMessenger Messenger
-		wantErrSubstr    string
+		name                string
+		notification        pubsub.Message
+		gcsObjectBytes      []byte
+		githubResourceBytes []byte
```
capri-xiyue, line 199: rename it.

pkg/server/event_handler_test.go
```diff
@@ -201,74 +220,100 @@ contacts:
...
-			successMessenger: &NoopMessenger{},
+			githubResourceBytes: getFakeMetadata(),
+			processors:          []Processor[*structpb.Struct]{&successProcessor{}},
+			successMessenger:    &NoopMessenger{},
 		},
 		{
 			name: "failed_send_downstream",
 			notification: pubsub.Message{
-				Attributes: map[string]string{"bucketId": "foo", "objectId": "bar"},
+				Attributes: map[string]string{"bucketId": "foo", "objectId": "bar", "git-commit": "test-git-commit"},
```
capri-xiyue, line 230: why do we need "git-commit" in the attributes?

test/integration/main_test.go
```diff
@@ -177,6 +186,10 @@ contacts:
 				t.Errorf("resourcemapping(ignore annotation) unexpected diff (-want,+got):\n%s", diff)
 			}
 
+			if diff := cmp.Diff(wantGithubResource, gotPmapEvent.GetGithubSource(), cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{})); diff != "" {
```
capri-xiyue, line 189: Do we have any unexported fileds in `GitHubSource`?

test/integration/main_test.go
```diff
@@ -214,6 +227,14 @@ func TestPolicyEventHandling(t *testing.T) {
 
 			ctx := context.Background()
 
+			wantGithubResource := &v1alpha1.GitHubSource{
```
capri-xiyue, line 230: GitHubSource instead of GitHubResource.

test/integration/main_test.go
```diff
@@ -262,6 +283,10 @@ deletion_timeline:
 			if diff := cmp.Diff(wantPayload, gotPayload, protocmp.Transform()); diff != "" {
 				t.Errorf("gotPayload unexpected diff (-want,+got):\n%s", diff)
 			}
+
+			if diff := cmp.Diff(wantGithubResource, gotPmapEvent.GetGithubSource(), cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{})); diff != "" {
```
capri-xiyue, line 287: same here

capri-xiyue (changes requested):

test/integration/main_test.go
```diff
@@ -129,6 +130,14 @@ func TestMappingEventHandling(t *testing.T) {
 	for _, tc := range cases {
 		tc := tc
 
+		wantGithubResource := &v1alpha1.GitHubSource{
```
capri-xiyue, line 133: same, rename

apis/v1alpha1/pmap_event.pb.go
```diff
@@ -123,6 +123,12 @@ type GitHubSource struct {
 	FilePath string `protobuf:"bytes,2,opt,name=file_path,json=filePath,proto3" json:"file_path,omitempty"`
 	// Required. The git commit.
 	Commit string `protobuf:"bytes,3,opt,name=commit,proto3" json:"commit,omitempty"`
+	// Required. The github workflow that triggered the pmap event.
+	Workflow string `protobuf:"bytes,4,opt,name=workflow,proto3" json:"workflow,omitempty"`
+	// Required. The sha for the github workflow.
+	WorkflowSha string `protobuf:"bytes,5,opt,name=workflow_sha,json=workflowSha,proto3" json:"workflow_sha,omitempty"`
+	// Required. The timestamp when workflow is triggered.
+	TriggeredTimestamp string `protobuf:"bytes,6,opt,name=triggered_timestamp,json=triggeredTimestamp,proto3" json:"triggered_timestamp,omitempty"`
```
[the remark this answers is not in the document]
sailorlqh: I used `protoc` to generate this file.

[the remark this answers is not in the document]
capri-xiyue: Can you use 'go-generate` https://github.com/abcxyz/pmap/blob/a69a6028c9f64ae936c19b883e2b1105546ca1fb/apis/v1alpha1/v1alpha1.go#L15? This is how we generate code in pmap

protos/v1alpha1/pmap_event.proto
sailorlqh: Change to time. Will do the renaming in the next PR along with all other renaming.

apis/v1alpha1/pmap_event.pb.go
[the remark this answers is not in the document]
capri-xiyue: I guess you need to update v1alpha1.go to include pmapevent

pkg/server/event_handler_test.go
sailorlqh: Removed. Forgot to remove those.

test/integration/main_test.go
sailorlqh: Yes, https://github.com/abcxyz/pmap/blob/add_pmap_event_info/apis/v1alpha1/pmap_event.pb.go#L116-L119

pkg/server/event_handler.go
```diff
@@ -293,6 +303,42 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
 	return p, nil
 }
 
+type payloadMetadata struct {
```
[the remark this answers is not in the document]
sailorlqh:
```
{
  "kind": "storage#object",
  "id": "some-id",
   ...
   ...
  "metadata": {
    (key): string
  },
  ...
  ...
}
```

pkg/server/event_handler_test.go
sailorlqh: pubsubMessageBytes is for test `invalid_request_body`, so we can directly return a [invalid request body](https://github.com/abcxyz/pmap/blob/add_pmap_event_info/pkg/server/event_handler_test.go#L124). 

pubSubMessage is for the other two tests, where we need an actual pubsubMessage.

pkg/server/event_handler_test.go
```diff
@@ -325,18 +370,40 @@ func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.
...
+		  "git-workflow-sha": "test-workflow-sha",
+		  "git-workflow": "test-workflow",
+		  "git-repo": "test-github-repo"
+		}
+	  }`)
+}
+
 // Returns a fake http func that writes the data in http response.
-func handleObjectRead(t *testing.T, data []byte) func(w http.ResponseWriter, r *http.Request) {
+func handleObjectRead(t *testing.T, data, metadata []byte) func(w http.ResponseWriter, r *http.Request) {
```
sqin2019, line 387: I am not sure why we need this func to handle object metadata, aren't the object metadata in notification payload? This func should only handles GCS return a object.

pkg/server/event_handler.go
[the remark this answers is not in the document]
sqin2019: will it work if the value is both string and a map?

pkg/server/event_handler.go
```diff
@@ -293,6 +305,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	pm := &payloadMetadata{}
+	if err := json.Unmarshal(data, pm); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payloadMetadata %w", err)
+	}
+
+	r := &v1alpha1.GitHubSource{}
+
+	c, found := pm.Metadata["git-commit"]
+	if found {
+		r.Commit = c
```
sqin2019, line 322: What we do if these are not found?

pkg/server/event_handler.go
```diff
@@ -293,6 +305,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	if found {
+		r.Workflow = w
+	}
+	ws, found := pm.Metadata["git-workflow-sha"]
+	if found {
+		r.WorkflowSha = ws
+	}
+	t, found := pm.Metadata["triggered-timestamp"]
+	if found {
+		date, err := time.Parse("2006-01-02T15:04:05", t)
```
sqin2019, line 339: Should we use time.RFC3339 for the format instead?

pkg/server/event_handler.go
```diff
@@ -293,6 +305,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+		if err != nil {
+			return nil, fmt.Errorf("error converting date %w", err)
```
sqin2019, line 341: Missing a unit test that throws this error?

pkg/server/event_handler_test.go
```diff
@@ -281,7 +326,7 @@ isOK: true`),
 			ctx := context.Background()
 
 			// Create fake http client for storage client.
-			hc, done := newTestServer(handleObjectRead(t, tc.gcsObjectBytes))
+			hc, done := newTestServer(handleObjectRead(t, tc.gcsObjectBytes, tc.githubSourceBytes))
```
sqin2019, line 329: Same to my other comments.

pkg/server/event_handler_test.go
sailorlqh: pubsubMessage.Message.Data has type `[]byte`. Initializing pubsub message would require we write code like 
```
PubSubMessageBytes:     []byte(`
{
"message": {
    "attributes": {
         "bucketId": "foo",
         "objectId": "bar2"
    },
    "data": [1 2 3 4 5 6 7 8]
}
`),
```
I think this is not good in terms of readability.

pkg/server/event_handler.go
sailorlqh: My idea is: if these are not found, than it means the user didn't provide these metadata. So we just ignore that.

test/integration/main_test.go
```diff
@@ -262,6 +295,14 @@ deletion_timeline:
 			if diff := cmp.Diff(wantPayload, gotPayload, protocmp.Transform()); diff != "" {
 				t.Errorf("gotPayload unexpected diff (-want,+got):\n%s", diff)
 			}
+			cmpOpts := []cmp.Option{
+				protocmp.Transform(),
+				protocmp.IgnoreFields(&v1alpha1.GitHubSource{}),
```
capri-xiyue, line 300: If you ignore this, what fields do you compare the diff in the `cmp.Diff(wantGithubSource, gotPmapEvent.GetGithubSource(), cmpOpts...)`

sailorlqh: Typo, I meant ignoreUnexported. Already updated the code.

pkg/server/event_handler.go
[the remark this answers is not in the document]
sailorlqh: In our current use case, the meta are all strings, I can update it to be able to accept map, just not sure it's necessary.

pkg/server/event_handler_test.go
yolocs: Should we just have `pubSubMessageBytes`? It's the raw input we would get from the HTTP request anyway.

pkg/server/event_handler_test.go
```diff
@@ -89,53 +90,79 @@ func TestEventHandler_HttpHandler(t *testing.T) {
...
-		name               string
-		pubsubMessageBytes []byte
-		gcsObjectBytes     []byte
-		wantStatusCode     int
-		wantRespBodySubstr string
+		name                   string
+		pubsubMessage          *PubSubMessage
+		PubSubMessageBytes     []byte
+		gcsObjectBytes         []byte
+		gcsObjectMetadataBytes []byte
```
yolocs, line 97: Do we need this if it's embedded in the pubsub message data?

pkg/server/event_handler_test.go
```diff
@@ -176,18 +211,20 @@ func TestEventHandler_Handle(t *testing.T) {
...
+		processors        []Processor[*structpb.Struct]
+		successMessenger  Messenger
+		failureMessenger  Messenger
+		wantErrSubstr     string
 	}{
 		{
 			name: "success",
 			notification: pubsub.Message{
 				Attributes: map[string]string{"bucketId": "foo", "objectId": "bar"},
+				Data:       getFakeMetadata(),
```
yolocs, line 227: This is added but I don't see we validate it in the result (in the successMessenger). Or we didn't do that in the first place, which we then should fix (could be a follow up)?

pkg/server/event_handler_test.go
```diff
@@ -176,18 +211,20 @@ func TestEventHandler_Handle(t *testing.T) {
...
-		notification     pubsub.Message
-		gcsObjectBytes   []byte
-		processors       []Processor[*structpb.Struct]
-		successMessenger Messenger
-		failureMessenger Messenger
-		wantErrSubstr    string
+		name              string
+		notification      pubsub.Message
+		gcsObjectBytes    []byte
+		githubSourceBytes []byte
```
yolocs, line 217: Not sure why it's `[]byte` because we are not validating it anywhere

pkg/server/event_handler.go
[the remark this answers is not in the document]
sqin2019: Based on https://cloud.google.com/storage/docs/json_api/v1/objects, it seems the data is a map of different values (strings, list, map, etc). how json.Unmarshal(data, map[string]string) behaves in that case?

pkg/server/event_handler_test.go
```diff
@@ -325,18 +388,40 @@ func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.
...
+			_, err := w.Write(data)
+			if err != nil {
+				t.Fatalf("failed to write response for object info: %v", err)
+			}
+		// This is for getting object's metadata
+		case "/storage/v1/b/foo/o/bar":
+			_, err := w.Write(getFakeMetadata())
+			if err != nil {
+				t.Fatalf("failed to write response for object metadata: %v", err)
+			}
```
yolocs, line 421: Do we still need this given we already get metadata from pubsub message?

yolocs (changes requested):

pkg/server/event_handler.go
sailorlqh: Currently the timestamp format is defined here: https://github.com/abcxyz/pmap/blob/main/.github/workflows/snapshot-file-change.yml#L35.

It has format of `2006-01-02T15:04:05`, but time.RFC3339 requires format of `2006-01-02T15:04:05Z07:00`

This requires modification to workflow files, I think we can fix this with #94

pkg/server/event_handler_test.go
sailorlqh: Discussed with @sqin2019, use pubSubMessageBytes for all tests now.

pkg/server/event_handler_test.go
sailorlqh: Created #97, will do this as a follow up.

pkg/server/event_handler_test.go
sailorlqh: Removed.

pkg/server/event_handler_test.go
sailorlqh: This is removed now, as we no longer needs this.

pkg/server/event_handler_test.go
sailorlqh: Removed, as it's no longer needed.

pkg/server/event_handler.go
```diff
@@ -228,9 +234,15 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 	if err != nil {
 		return fmt.Errorf("failed to convert object to pmap event payload: %w", err)
 	}
-	// TODO(#21): Add additional metadata to pmap event.
+
```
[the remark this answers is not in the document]
sailorlqh: Added. https://github.com/abcxyz/pmap/blob/add_pmap_event_info/pkg/server/event_handler.go#L239-L244

pkg/server/event_handler.go
```diff
@@ -161,6 +166,7 @@ type PubSubMessage struct {
 // in HTTP requests and calls [Handle] to handle the events.
 //
 // [GCS notifications]: https://cloud.google.com/storage/docs/pubsub-notifications#format
+// Object's metadata change will be included in payload of the notification.
```
yolocs, line 169: nit: Can you keep the links at the end of comment?
```suggestion
// in HTTP requests and calls [Handle] to handle the events.
// Object's metadata change will be included in payload of the notification.
//
// [GCS notifications]: https://cloud.google.com/storage/docs/pubsub-notifications#format
```

pkg/server/event_handler.go
```diff
@@ -228,9 +234,18 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 	if err != nil {
 		return fmt.Errorf("failed to convert object to pmap event payload: %w", err)
 	}
-	// TODO(#21): Add additional metadata to pmap event.
+
+	var gr *v1alpha1.GitHubSource
+	if m.Attributes["payloadFormat"] == "JSON_API_V1" {
```
yolocs, line 239: What if it's not `JSON_API_V1`?

pkg/server/event_handler_test.go
yolocs: If it's not used, should we defer adding this?

pkg/server/event_handler_test.go
```diff
@@ -325,18 +404,34 @@ func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.
 	}
 }
 
+// Returns fake metadata that include github resource info.
+func getFakeMetadata() []byte {
```
yolocs, line 408: https://github.com/abcxyz/readability/wiki/Go-Best-Practices#test-helper
```suggestion
func testGCSMetadataBytes() []byte {
```

pkg/server/event_handler_test.go
```diff
@@ -325,18 +404,34 @@ func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.
...
+		  "triggered-timestamp": "2023-04-25T17:44:57",
+		  "git-workflow-sha": "test-workflow-sha",
+		  "git-workflow": "test-workflow",
+		  "git-repo": "test-github-repo"
+		}
+	  }`)
+}
+
 // Returns a fake http func that writes the data in http response.
 func handleObjectRead(t *testing.T, data []byte) func(w http.ResponseWriter, r *http.Request) {
```
yolocs, line 421: https://github.com/abcxyz/readability/wiki/Go-Best-Practices#test-helper
```suggestion
func testHandleObjectRead(t *testing.T, data []byte) func(w http.ResponseWriter, r *http.Request) {
```

pkg/server/event_handler.go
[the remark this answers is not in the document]
yolocs: Not resolved

pkg/server/event_handler.go
```diff
@@ -293,6 +305,46 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+
+func parseGitHubSource(ctx context.Context, data []byte) (*v1alpha1.GitHubSource, error) {
+	pm := &payloadMetadata{}
+	if err := json.Unmarshal(data, pm); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payloadMetadata %w", err)
+	}
+
+	r := &v1alpha1.GitHubSource{}
+
+	c, found := pm.Metadata["git-commit"]
```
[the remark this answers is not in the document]
yolocs: Not resolved

pkg/server/event_handler_test.go
```diff
@@ -97,45 +98,87 @@ func TestEventHandler_HttpHandler(t *testing.T) {
...
+				},
+				"data" : %q
+			  },
+			  "subscription": "test_subscription"
+			}
+			`, base64.StdEncoding.EncodeToString([]byte(`{
+				"metadata": {
+					""
+				}
+			  }`)))),
```
yolocs, line 178: You can direct replace %q with a random string. It'd be more obvious it's invalid.

yolocs (changes requested):

pkg/server/event_handler.go
sailorlqh: If it's not `JSON_API_V1`, the payload would be `None`. In this case we don't need to parseGithubSource.
https://cloud.google.com/storage/docs/pubsub-notifications#payload

pkg/server/event_handler.go
sailorlqh: Added to use time.RFC3339

pkg/server/event_handler_test.go
sailorlqh: data's value needs to be of type `[]byte`, putting a string there will cause problem for handler to parse the response body into `PubsubMessage`. Use %q here is just to bypass that.

pkg/server/event_handler_test.go
sailorlqh: `Data` in `notification` has the githubSourceMetadata which will be used in [parseGitHubSource](https://github.com/abcxyz/pmap/blob/add_pmap_event_info/pkg/server/event_handler.go#L324).

pkg/server/event_handler_test.go
sethvargo: @yolocs `[]byte` is base64 encoded in JSON.

@sailorlqh it would be slightly better to use a real structure here instead of compiling a string. If you need to compile a string, using "%s" would be less indirect than %q.

sailorlqh: Updated it to a real structure.

yolocs: Oh right I missed that.

pkg/server/event_handler.go
```diff
@@ -55,6 +57,14 @@ type Processor[P proto.Message] interface {
 	Stop() error
 }
 
+const (
+	MetadataKeyGitHubCommit       = "git-commit"
```
sethvargo, line 61: It could be helpful to document these constants (and where they come from).

pkg/server/event_handler.go
```diff
@@ -293,6 +317,44 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	Metadata map[string]string `json:"metadata,omitempty"`
+}
+
+func parseGitHubSource(ctx context.Context, data []byte) (*v1alpha1.GitHubSource, error) {
+	var pm payloadMetadata
+	if err := json.Unmarshal(data, &pm); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payloadMetadata %w", err)
+	}
+
+	r := &v1alpha1.GitHubSource{}
```
sethvargo, line 330: Nit: It would be more idiomatic to do:

```go
var r v1alpha1.GitHubSource
```

since there are no fields being set on the struct during initialization.

pkg/server/event_handler.go
```diff
@@ -293,6 +317,44 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	}
+
+	if ws, found := pm.Metadata[MetadataKeyWorkflowSha]; found {
+		r.WorkflowSha = ws
+	}
+
+	if t, found := pm.Metadata[MetadataKeyTriggeredTimestamp]; found {
+		date, err := time.Parse(time.RFC3339, t)
+		if err != nil {
+			return nil, fmt.Errorf("failed converting date %w", err)
```
sethvargo, line 351:
```suggestion
			return nil, fmt.Errorf("failed to parse date: %w", err)
```

pkg/server/event_handler_test.go
```diff
@@ -97,45 +98,87 @@ func TestEventHandler_HttpHandler(t *testing.T) {
 	}{
 		{
 			name: "success",
-			pubsubMessageBytes: []byte(`
+			pubsubMessageBytes: []byte(fmt.Sprintf(`
```
sethvargo, line 101: I think it would be clearer if this was a dedicated struct with a helper to serialize it, like:

```go
pubsubMessageBytes: toJSON(t, &MyStructure{
  Message: &Message{
    Attributes: &Attributes{},
  },
})

// ...

func testToJSON(tb testing.TB, in any)[]byte {
  tb.Helper()

  b, err := json.Marshal(in)
  if err != nil {
    tb.Fatal(err)
  }

  return b
}
```

If using a struct is too heavy, you can use a `map[any]any`:

```go
pubsubMessageBytes: toJSON(t, map[string]any{
  "message": map[string]any{
    "attributes": map[string]any{
      "bucketId": "foo",
    },
    "data": []byte(`{...}`)
  },
})
```

You don't have to b64 encode the data if it's a structure and `[]byte`.

protos/v1alpha1/pmap_event.proto
```diff
@@ -44,4 +44,13 @@ message GitHubSource {
 
   // Required. The git commit.
   string commit = 3;
+
+  // Required. The github workflow that triggered the pmap event.
```
sethvargo, line 48: Please provide an example of the value.

protos/v1alpha1/pmap_event.proto
```diff
@@ -44,4 +44,13 @@ message GitHubSource {
...
+  string workflow = 4;
+
+  // Required. The sha for the github workflow.
```
sethvargo, line 51: Please provide an example of the value.

test/integration/main_test.go
```diff
@@ -128,6 +131,18 @@ func TestMappingEventHandling(t *testing.T) {
 
 	for _, tc := range cases {
 		tc := tc
+		wantTS := "2023-04-25T17:44:57Z"
+		wantT, err := time.Parse(time.RFC3339, wantTS)
+		if err != nil {
+			t.Fatalf("failed to parse timestamp str to timestamppb")
```
sethvargo, line 137:
```suggestion
			t.Fatal(err)
```

pkg/server/event_handler_test.go
sailorlqh: Updated.

test/integration/main_test.go
```diff
@@ -128,6 +131,18 @@ func TestMappingEventHandling(t *testing.T) {
...
+		}
+		wantGithubSource := &v1alpha1.GitHubSource{
+			RepoName:           "test-github-repo",
+			Commit:             "test-github-commit",
+			Workflow:           "test-workflow",
+			WorkflowSha:        "test-workflow-sha",
+			TriggeredTimestamp: timestamppb.New(wantT),
+		}
```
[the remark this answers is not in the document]
sailorlqh: Moved to sub-tests.

pkg/server/event_handler_test.go
```diff
@@ -97,45 +98,87 @@ func TestEventHandler_HttpHandler(t *testing.T) {
...
-			{
-				"message": {
-					"attributes": {
-						"bucketId": "foo",
-						"objectId": "bar2"
-					}
+			pubsubMessageBytes: testToJSON(t, &PubSubMessage{
+				Message: struct {
+					Data       []byte            "json:\"data,omitempty\""
+					Attributes map[string]string "json:\"attributes\""
```
sethvargo, line 139:
```suggestion
					Data       []byte            `json:"data,omitempty"`
					Attributes map[string]string `json:"attributes"`
```

pkg/server/event_handler_test.go
```diff
@@ -97,45 +98,87 @@ func TestEventHandler_HttpHandler(t *testing.T) {
...
+			}),
 			wantStatusCode:     http.StatusInternalServerError,
 			wantRespBodySubstr: "failed to get GCS object",
 		},
+		{
+			name: "invalid_pubsubmessage_data",
+			pubsubMessageBytes: testToJSON(t, &PubSubMessage{
+				Message: struct {
+					Data       []byte            "json:\"data,omitempty\""
+					Attributes map[string]string "json:\"attributes\""
```
sethvargo, line 165:
```suggestion
					Data       []byte            `json:"data,omitempty"`
					Attributes map[string]string `json:"attributes"`
```

pkg/server/event_handler_test.go
```diff
@@ -176,18 +218,20 @@ func TestEventHandler_Handle(t *testing.T) {
...
 	cases := []struct {
-		name             string
-		notification     pubsub.Message
-		gcsObjectBytes   []byte
-		processors       []Processor[*structpb.Struct]
-		successMessenger Messenger
-		failureMessenger Messenger
-		wantErrSubstr    string
+		name              string
+		notification      pubsub.Message
```
sethvargo, line 222: Nit: this should be `*pubsub.Message`.

protos/v1alpha1/pmap_event.proto
sethvargo: I'm not seeing where this was added.

test/integration/main_test.go
```diff
@@ -362,3 +417,12 @@ func testUploadFile(ctx context.Context, tb testing.TB, bucket, object string, d
 
 	return nil
 }
+
+func testParseTime(tb testing.TB, ts string) *timestamppb.Timestamp {
+	tb.Helper()
+	t, err := time.Parse(time.RFC3339, ts)
+	if err != nil {
+		tb.Fatal("failed to parse timestamp string to time")
```
sethvargo, line 425:
```suggestion
		tb.Fatal(err)
```

pkg/server/event_handler.go
```diff
@@ -55,6 +57,18 @@ type Processor[P proto.Message] interface {
...
+// These are metadatas for GCS objects that were uploaded.
+// These customs keys are defined in snapshot-file-change
+// and snapshot-file-copy workflow.
+// https://github.com/abcxyz/pmap/blob/main/.github/workflows/snapshot-file-change.yml#L74-L78
+const (
+	MetadataKeyGitHubCommit               = "git-commit"
+	MetadataKeyGitHubRepo                 = "git-repo"
+	MetadataKeyWorkflow                   = "git-workflow"
+	MetadataKeyWorkflowSha                = "git-workflow-sha"
+	MetadataKeyWorkflowTriggeredTimestamp = "workflow-triggered-timestamp"
```
yolocs, line 69: Mind prefixing it with `git-`?

pkg/server/event_handler_test.go
```diff
@@ -97,45 +98,87 @@ func TestEventHandler_HttpHandler(t *testing.T) {
...
-			{
-				"message": {
-					"attributes": {
-						"bucketId": "foo",
-						"objectId": "bar"
-					}
+			pubsubMessageBytes: testToJSON(t, &PubSubMessage{
+				Message: struct {
+					Data       []byte            "json:\"data,omitempty\""
+					Attributes map[string]string "json:\"attributes\""
```
yolocs, line 104: Fix tags

protos/v1alpha1/pmap_event.proto
sailorlqh: @sethvargo this will be added when fixing #94.

yolocs (approved):

pkg/server/event_handler.go
```diff
@@ -293,6 +321,44 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
 	return p, nil
 }
 
+type payloadMetadata struct {
```
capri-xiyue, line 324: rename to notificationPayload?

pkg/server/event_handler.go
```diff
@@ -293,6 +321,44 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
...
+	Metadata map[string]string `json:"metadata,omitempty"`
+}
+
+func parseGitHubSource(ctx context.Context, data []byte) (*v1alpha1.GitHubSource, error) {
+	var pm payloadMetadata
+	if err := json.Unmarshal(data, &pm); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payloadMetadata %w", err)
+	}
+
+	var r v1alpha1.GitHubSource
```
capri-xiyue, line 334: use pointer here

capri-xiyue (approved):

protos/v1alpha1/pmap_event.proto
sethvargo: ack

.github/workflows/snapshot-file-change.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           fetch-depth: 0 # OR "2" -> To retrieve the preceding commit.
       - name: 'Set current timestamp'
         run: |
-          echo "CUR_TIMESTAMP=$(date +'%Y-%m-%dT%H:%M:%S')" >> $GITHUB_ENV
+          echo "CUR_TIMESTAMP=$(date +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_ENV
```
sqin2019, line 35: Not sure the Z here is right, how about change to 
```
echo "CUR_TIMESTAMP=$(date --rfc-3339=seconds | sed 's/ /T/')"
```

sailorlqh: Will do this together with #94

sqin2019 (approved):

PR #104, opened 2023-05-30
add_verbosity

sailorlqh:


pkg/cli/mapping_validate.go
```diff
@@ -114,7 +125,7 @@ func (c *MappingValidateCommand) validateResourceMappings() error {
 			continue
 		}
 	}
-	if checkErrs == nil {
+	if checkErrs == nil && c.verbosity {
```
capri-xiyue, line 128: validation pass or not should be an output all the time

pkg/cli/mapping_validate.go
```diff
@@ -33,6 +33,8 @@ type MappingValidateCommand struct {
 	cli.BaseCommand
 
 	flagPath string
+
+	verbosity bool
```
drevell, line 37: I think `verbose` would be a better name for a bool than `verbosity`

drevell (approved):

pkg/cli/mapping_validate.go
yolocs, line 37: I think we should standardize on `log-level` rather than introducing another flag and leave it to the logger. See https://github.com/abcxyz/pkg/issues/121. We decided to punt this until we introduce `slog` package. Maybe we should punt this issue in pmap as well.

sailorlqh:
Close this PR, will wait until https://github.com/abcxyz/pkg/issues/121 is fixed, and standardize on `log-level`.

Issue #97, opened 2023-05-16
Validate githubSource metadata in the result (in the successMessenger)


sailorlqh:
https://github.com/abcxyz/pmap/pull/93#discussion_r1195694682

PR #105, opened 2023-05-31
Fix: update test to validate pmapEvent in successMessenger

sailorlqh:
fix: #97

### DIFF
--- a/pkg/server/event_handler_test.go
+++ b/pkg/server/event_handler_test.go
@@ -25,13 +25,18 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"cloud.google.com/go/pubsub"
 	"cloud.google.com/go/storage"
 	"github.com/abcxyz/pkg/testutil"
 	"github.com/abcxyz/pmap/apis/v1alpha1"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"google.golang.org/api/option"
+	"google.golang.org/protobuf/testing/protocmp"
 	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 func TestEventHandler_NewHandler(t *testing.T) {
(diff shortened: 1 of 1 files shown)


sethvargo (approved):

pkg/server/event_handler_test.go
```diff
@@ -468,12 +498,42 @@ func (p *successProcessor) Process(_ context.Context, m *structpb.Struct) error
 }
 
 // failMessenger Send always fail.
-type failMessenger struct{}
 
-func (m *failMessenger) Send(_ context.Context, _ *v1alpha1.PmapEvent) error {
+type testMessanger interface {
+	Messenger
+	getPmapEvent() *v1alpha1.PmapEvent
+}
```
yolocs, line 505: We don't need this interface as long as our fake messenger implements the `Messenger` interface.

pkg/server/event_handler_test.go
```diff
@@ -468,12 +498,42 @@ func (p *successProcessor) Process(_ context.Context, m *structpb.Struct) error
...
+
+type failMessenger struct {
+	gotPmapEvent *v1alpha1.PmapEvent
+}
```
yolocs, line 509: Can we just have one fake messenger? The only difference between the two is one requires error. You achieve the same with a field. E.g.

```go
type testMessenger struct {
  returnErr error
  gotEvent *v1alpha1.PmapEvent
}

func (m *testMessenger) Send(_ context.Context, p *v1alpha1.PmapEvent) error {
  m.gotEvent = p
  return m.returnErr
}

// We should get rid of these in a separate refactor
// See https://github.com/abcxyz/pmap/issues/96
func (m *testMessenger) Cleanup() error {
	return nil
}
```

yolocs (changes requested):

yolocs (approved):

pkg/server/event_handler_test.go
```diff
@@ -388,6 +427,18 @@ isOK: true`),
...
 				t.Errorf("Process(%+v) got unexpected error substring: %v", tc.name, diff)
 			}
+
+			cmpOpts := []cmp.Option{
+				protocmp.Transform(),
+				protocmp.IgnoreFields(&v1alpha1.PmapEvent{}, "payload"),
+				cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{}),
+				cmpopts.IgnoreUnexported(timestamppb.Timestamp{}),
+			}
+			if tc.wantPmapEvent != nil {
```
yolocs, line 437: I think you don't need this `if`. cmp.Diff can handle nil

pkg/server/event_handler_test.go
```diff
@@ -388,6 +446,16 @@ isOK: true`),
 			if diff := testutil.DiffErrString(gotErr, tc.wantErrSubstr); diff != "" {
 				t.Errorf("Process(%+v) got unexpected error substring: %v", tc.name, diff)
 			}
+
+			cmpOpts := []cmp.Option{
+				protocmp.Transform(),
+				protocmp.IgnoreFields(&v1alpha1.PmapEvent{}, "payload"),
+				cmpopts.IgnoreUnexported(v1alpha1.GitHubSource{}),
```
capri-xiyue, line 453: Why do we need these two  IgnoreUnexported?

sailorlqh: Forgot to remove it. Just removed.

capri-xiyue (approved):

PR #107, opened 2023-06-01
Install yamllint

sethvargo:
This will fail until https://github.com/abcxyz/pkg/pull/132 is merged.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,7 +25,7 @@ jobs:
           files: |
             ${{ inputs.resource_mapping_directory }}/**/*.{yaml,yml}
       - name: 'Copy added and changed files'
-        if: ${{ steps.changed-yaml-files.outputs.any_changed == 'true' }}
+        if: '${{ steps.changed-yaml-files.outputs.any_changed == ''true'' }}'
         run: |
           echo ${{ env.RUNNER_TEMP }}
           sudo mkdir -p ${{ env.RUNNER_TEMP }}/pmap/snapshot-file-change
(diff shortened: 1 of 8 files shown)


capri-xiyue (approved):

.github/workflows/check.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           files: |
             ${{ inputs.resource_mapping_directory }}/**/*.{yaml,yml}
       - name: 'Copy added and changed files'
-        if: ${{ steps.changed-yaml-files.outputs.any_changed == 'true' }}
+        if: '${{ steps.changed-yaml-files.outputs.any_changed == ''true'' }}'
```
sqin2019, line 28: @sethvargo Hi, with this, vscode IDE gives an error even though workflows do not complain, wondering what setting in the IDE can fix it.
```
Unexpected symbol: 'true'''. Located at position 30 within expression: needs.check.outputs.aod == ''true''
```

sethvargo: Hey @sqin2019 - hmm that feels like a buge in the editor. Which plugin is throwing the error?

sqin2019: Not sure, will check, so no special IDE settings are needed for this I assume.

sqin2019: seems like it is the github action extention https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions

sethvargo: Got it - it feels like an extension of https://github.com/github/vscode-github-actions/issues/60? Do you mind filing an issue there?

sqin2019: https://github.com/github/vscode-github-actions/issues/211

Issue #96, opened 2023-05-16
Let caller handle client lifecycle


sqin2019:
Follow up of https://github.com/abcxyz/pmap/pull/89#issuecomment-1548693319

PR #115, opened 2023-06-07
refactor: refactore messanger interface

sailorlqh:
Refactor the code as a prerequisite for [PR#114](https://github.com/abcxyz/pmap/pull/114), as discussed here: https://github.com/abcxyz/pmap/pull/114#issuecomment-1581375884.

Refactor the Messenger's Send() interface to accept data as byte and attributes as map, as they are needed for publishing the message.

Also removed Cleanup() from Messenger's interface, and move cleanup logic to be handled outside of handler.

fix: #96.

### DIFF
--- a/pkg/cli/mapping_server.go
+++ b/pkg/cli/mapping_server.go
@@ -117,12 +117,6 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 		return nil, nil, closer, fmt.Errorf("server.NewHandler: %w", err)
 	}
 
-	closer = func() {
-		if err := handler.Cleanup(); err != nil {
-			logger.Errorw("failed to close clean up handler", "error", err)
-		}
-	}
-
 	srv, err := serving.New(c.cfg.Port)
 	if err != nil {
 		return nil, nil, closer, fmt.Errorf("failed to create serving infrastructure: %w", err)
--- a/pkg/cli/mapping_server.go
+++ b/pkg/cli/mapping_server.go
@@ -98,17 +98,14 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 		return nil, nil, closer, fmt.Errorf("missing PMAP_FAILURE_TOPIC_ID in config")
 	}
 
-	pubsubClient, err := createPubsubClient(ctx, c.cfg.ProjectID)
+	pubsubClient, err := pubsub.NewClient(ctx, c.cfg.ProjectID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create pubsub client: %w", err)
 	}
 
 	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
-
 	successMessenger := server.NewPubSubMessenger(successTopic)
-
 	failureTopic := pubsubClient.Topic(c.cfg.FailureTopicID)
-
 	failureMessenger := server.NewPubSubMessenger(failureTopic)
 
 	assetClient, err := asset.NewClient(ctx)
@@ -129,11 +126,6 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 		return nil, nil, closer, fmt.Errorf("server.NewHandler: %w", err)
 	}
 
-	srv, err := serving.New(c.cfg.Port)
-	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create serving infrastructure: %w", err)
-	}
-
 	closer = func() {
 		successTopic.Stop()
 		failureTopic.Stop()
@@ -147,13 +139,10 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 		}
 	}
 
-	return srv, handler.HTTPHandler(), closer, nil
-}
-
-func createPubsubClient(ctx context.Context, projectID string) (*pubsub.Client, error) {
-	client, err := pubsub.NewClient(ctx, projectID)
+	srv, err := serving.New(c.cfg.Port)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create new pubsub client: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create serving infrastructure: %w", err)
 	}
-	return client, nil
+
+	return srv, handler.HTTPHandler(), closer, nil
 }
--- a/pkg/cli/mapping_server.go
+++ b/pkg/cli/mapping_server.go
@@ -16,7 +16,6 @@ package cli
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 
@@ -66,6 +65,7 @@ func (c *MappingServerCommand) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
+
 	defer closer()
 
 	return srv.StartHTTPHandler(ctx, handler)
@@ -97,28 +97,28 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 	if c.cfg.FailureTopicID == "" {
 		return nil, nil, closer, fmt.Errorf("missing PMAP_FAILURE_TOPIC_ID in config")
 	}
-	successTopic, err := createMessangerTopic(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
+
+	pubsubClient, err := createPubsubClient(ctx, c.cfg.ProjectID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client and topic: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client: %w", err)
 	}
+
+	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
 
 	successMessenger := server.NewPubSubMessenger(successTopic)
 
-	failureTopic, err := createMessangerTopic(ctx, c.cfg.ProjectID, c.cfg.FailureTopicID)
-	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create failure event pubsub client and topic: %w", err)
-	}
+	failureTopic := pubsubClient.Topic(c.cfg.FailureTopicID)
 
 	failureMessenger := server.NewPubSubMessenger(failureTopic)
 
-	processorClient, err := asset.NewClient(ctx)
+	assetClient, err := asset.NewClient(ctx)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create the Asset Inventory client: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create the assetClient: %w", err)
 	}
 
-	processor, err := processors.NewAssetInventoryProcessor(ctx, processorClient, fmt.Sprintf("projects/%s", c.cfg.ProjectID))
+	processor, err := processors.NewAssetInventoryProcessor(ctx, assetClient, fmt.Sprintf("projects/%s", c.cfg.ProjectID))
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create asset inventory processor: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create assetInventoryProcessor: %w", err)
 	}
 
 	handler, err := server.NewHandler(ctx,
@@ -135,25 +135,25 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 	}
 
 	closer = func() {
-		var retErr error
 		successTopic.Stop()
 		failureTopic.Stop()
-		if err := processorClient.Close(); err != nil {
-			retErr = errors.Join(retErr, fmt.Errorf("failed to stop processor: %w", err))
+
+		if err := pubsubClient.Close(); err != nil {
+			logger.Errorw("failed to close pubsub client", "error", err)
 		}
-		if retErr != nil {
-			logger.Errorw("failed to clean up resources", "error", err)
+
+		if err := assetClient.Close(); err != nil {
+			logger.Errorw("failed to stop asset client", "error", err)
 		}
 	}
 
 	return srv, handler.HTTPHandler(), closer, nil
 }
 
-func createMessangerTopic(ctx context.Context, projectID, topicID string) (*pubsub.Topic, error) {
+func createPubsubClient(ctx context.Context, projectID string) (*pubsub.Client, error) {
 	client, err := pubsub.NewClient(ctx, projectID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new pubsub client: %w", err)
 	}
-	topic := client.Topic(topicID)
-	return topic, nil
+	return client, nil
 }
--- a/pkg/cli/mapping_server.go
+++ b/pkg/cli/mapping_server.go
@@ -20,7 +20,7 @@ import (
 	"fmt"
 	"net/http"
 
-	asset "cloud.google.com/go/asset/apiv1"
+	"cloud.google.com/go/pubsub"
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/serving"
@@ -28,6 +28,9 @@ import (
 	"github.com/abcxyz/pmap/internal/version"
 	"github.com/abcxyz/pmap/pkg/mapping/processors"
 	"github.com/abcxyz/pmap/pkg/server"
+	"google.golang.org/api/option"
+
+	asset "cloud.google.com/go/asset/apiv1"
 )
 
 var _ cli.Command = (*MappingServerCommand)(nil)
@@ -36,6 +39,14 @@ type MappingServerCommand struct {
 	cli.BaseCommand
 
 	cfg *server.HandlerConfig
+
+	successTopic  *pubsub.Topic
+	successClient *pubsub.Client
+
+	failureTopic  *pubsub.Topic
+	failureClient *pubsub.Client
+
+	processorClient *asset.Client
 
 	// testFlagSetOpts is only used for testing.
 	testFlagSetOpts []cli.Option
@@ -64,9 +75,6 @@ func (c *MappingServerCommand) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return err
 	}
-	// defer func() {
-	//     err = logger.Sync()
-	// }()
 	defer closer()
 
 	return srv.StartHTTPHandler(ctx, handler)
@@ -98,23 +106,27 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string) 
 	if c.cfg.FailureTopicID == "" {
 		return nil, nil, closer, fmt.Errorf("missing PMAP_FAILURE_TOPIC_ID in config")
 	}
-
-	successMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
+	var err error
+	c.successClient, c.successTopic, err = createMessangerClientAndTopic(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event messenger: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client and topic: %w", err)
 	}
 
-	failureMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.FailureTopicID)
+	successMessenger := server.NewPubSubMessenger(c.successClient, c.successTopic)
+
+	c.failureClient, c.failureTopic, err = createMessangerClientAndTopic(ctx, c.cfg.ProjectID, c.cfg.FailureTopicID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create failure event messenger: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create failure event pubsub client and topic: %w", err)
 	}
 
-	client, err := asset.NewClient(ctx)
+	failureMessenger := server.NewPubSubMessenger(c.failureClient, c.failureTopic)
+
+	c.processorClient, err = asset.NewClient(ctx)
 	if err != nil {
 		return nil, nil, closer, fmt.Errorf("failed to create the Asset Inventory client: %w", err)
 	}
 
-	processor, err := processors.NewAssetInventoryProcessor(ctx, fmt.Sprintf("projects/%s", c.cfg.ProjectID), client)
+	processor, err := processors.NewAssetInventoryProcessor(ctx, c.processorClient, fmt.Sprintf("projects/%s", c.cfg.ProjectID))
 	if err != nil {
 		return nil, nil, closer, fmt.Errorf("failed to create asset inventory processor: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


pkg/mapping/processors/assetinventory.go
```diff
@@ -60,7 +60,7 @@ func WithClient(client *asset.Client) Option {
 // NewAssetInventoryProcessor creates a new AssetInventoryProcessor with the given options.
 // Need defaultResourceScope because resources such as GCS bucket won't include Project info in its resource name.
 // See details: https://cloud.google.com/asset-inventory/docs/resource-name-format.
-func NewAssetInventoryProcessor(ctx context.Context, defaultResourceScope string, opts ...Option) (*AssetInventoryProcessor, error) {
+func NewAssetInventoryProcessor(ctx context.Context, defaultResourceScope string, client *asset.Client, opts ...Option) (*AssetInventoryProcessor, error) {
```
sqin2019, line 63: Would prefer  func NewAssetInventoryProcessor(ctx context.Context, client *asset.Client, defaultResourceScope string, opts ...Option)

pkg/mapping/processors/assetinventory.go
```diff
@@ -60,7 +60,7 @@ func WithClient(client *asset.Client) Option {
 // NewAssetInventoryProcessor creates a new AssetInventoryProcessor with the given options.
```
sqin2019, line 60: We should remove below func in this file
```
func (p *AssetInventoryProcessor) Stop() error {
```

pkg/server/event_handler.go
```diff
@@ -80,8 +80,7 @@ const (
 
 // An interface for sending pmap event downstream.
 type Messenger interface {
-	Send(context.Context, *v1alpha1.PmapEvent) error
-	Cleanup() error
+	Send(context.Context, []byte, map[string]string) error
```
sqin2019, line 83: Since we removed cleanup for messenger, we should do the same, which is when creating the messenger, instead of creating the clients inside the messenger [here](https://github.com/abcxyz/pmap/blob/bca64a43b358d27f9613407a2858bc788eacf88a/pkg/server/pubsub.go#L35), pass it directly.

pkg/cli/mapping_server.go
```diff
@@ -104,7 +108,12 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
 		return nil, nil, closer, fmt.Errorf("failed to create failure event messenger: %w", err)
 	}
 
-	processor, err := processors.NewAssetInventoryProcessor(ctx, fmt.Sprintf("projects/%s", c.cfg.ProjectID))
+	client, err := asset.NewClient(ctx)
```
sqin2019, line 111: This client needs to be closed?

pkg/cli/mapping_server.go
```diff
@@ -117,15 +126,24 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+		var retErr error
+		if err := successMessenger.Cleanup(); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to close success event messenger: %w", err))
+		}
+		if err := failureMessenger.Cleanup(); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to close failure event messenger: %w", err))
+		}
+		if err := processor.Stop(); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to stop processor: %w", err))
+		}
```
sqin2019, line 144: Our goal is to remove all this func, and call the client directly to close them here.

pkg/server/pubsub.go
```diff
@@ -31,39 +28,19 @@ type PubSubMessenger struct {
...
-func NewPubSubMessenger(ctx context.Context, projectID, topicID string, opts ...option.ClientOption) (*PubSubMessenger, error) {
-	client, err := pubsub.NewClient(ctx, projectID, opts...)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create new pubsub client: %w", err)
-	}
-
-	topic := client.Topic(topicID)
-
-	return &PubSubMessenger{client: client, topic: topic}, nil
+func NewPubSubMessenger(client *pubsub.Client, topic *pubsub.Topic) *PubSubMessenger {
```
yolocs, line 31: It seems we only need the topic. We don't need a client here.

pkg/server/event_handler.go
```diff
@@ -270,37 +269,25 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 		GithubSource: gr,
 	}
 
+	eventByte, err := parsePmapEvent(event)
+	if err != nil {
+		return err
+	}
+
 	if processErr != nil {
+		attr := map[string]string{}
```
yolocs, line 278: Define this outside of the if condition and use it in successMessenger call

pkg/server/event_handler.go
```diff
@@ -336,6 +323,14 @@ type notificationPayload struct {
 	Metadata map[string]string `json:"metadata,omitempty"`
 }
 
+func parsePmapEvent(event *v1alpha1.PmapEvent) ([]byte, error) {
+	eventByte, err := protojson.Marshal(event)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal event to byte: %w", err)
+	}
+	return eventByte, nil
+}
```
yolocs, line 332: Don't see the point of having this helper. You're introducing more lines than w/o it

pkg/cli/mapping_server.go
```diff
@@ -35,6 +40,14 @@ type MappingServerCommand struct {
 
 	cfg *server.HandlerConfig
 
+	successTopic  *pubsub.Topic
+	successClient *pubsub.Client
+
+	failureTopic  *pubsub.Topic
+	failureClient *pubsub.Client
+
+	processorClient *asset.Client
```
yolocs, line 49: Why make them fields?

pkg/cli/mapping_server.go
```diff
@@ -117,15 +139,40 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create serving infrastructure: %w", err)
+		return nil, nil, fmt.Errorf("failed to create new pubsub client: %w", err)
 	}
-	return srv, handler.HTTPHandler(), closer, nil
+	topic := client.Topic(topicID)
+	return client, topic, nil
+}
+
+func (c *MappingServerCommand) Cleanup() (retErr error) {
```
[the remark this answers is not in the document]
yolocs: Plus, should be able to inline these directly in the `closer` func above.

yolocs (changes requested):

pkg/cli/mapping_server.go
```diff
@@ -117,15 +129,31 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+			logger.Errorw("failed to clean up resources", "error", err)
 		}
 	}
 
-	srv, err := serving.New(c.cfg.Port)
+	return srv, handler.HTTPHandler(), closer, nil
+}
+
+func createMessangerTopic(ctx context.Context, projectID, topicID string) (*pubsub.Topic, error) {
+	client, err := pubsub.NewClient(ctx, projectID)
```
yolocs, line 153: We can create a single pubsub client since we only have a single project. You can use a single client to create two topics. I think we can get rid of this func

pkg/cli/mapping_server.go
```diff
@@ -93,18 +97,26 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
 	if c.cfg.FailureTopicID == "" {
 		return nil, nil, closer, fmt.Errorf("missing PMAP_FAILURE_TOPIC_ID in config")
 	}
+	successTopic, err := createMessangerTopic(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
```
yolocs, line 100: Nit: add an empty line above

pkg/cli/mapping_server.go
```diff
@@ -117,15 +129,31 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
-			logger.Errorw("failed to close clean up handler", "error", err)
+		var retErr error
+		successTopic.Stop()
+		failureTopic.Stop()
+		if err := processorClient.Close(); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to stop processor: %w", err))
+		}
+		if retErr != nil {
+			logger.Errorw("failed to clean up resources", "error", err)
 		}
```
yolocs, line 146: If we return the err, it will be logged. No need to log here.

pkg/server/pubsub_test.go
```diff
@@ -106,3 +114,15 @@ func newTestPubSubGrpcConn(ctx context.Context, t *testing.T, opts ...pstest.Ser
 
 	return conn
 }
+
+func createTestMessangerTopic(ctx context.Context, projectID, topicID string, opts ...option.ClientOption) (*pubsub.Topic, error) {
```
yolocs, line 118: Errors should be handled directly in the test helper, see https://google.github.io/styleguide/go/best-practices#test-helper-error-handling
Test helper should be named `testXxx` https://github.com/abcxyz/readability/wiki/Go-Best-Practices#test-helper

pkg/server/event_handler.go
```diff
@@ -270,37 +269,26 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 		GithubSource: gr,
 	}
 
+	eventByte, err := protojson.Marshal(event)
```
yolocs, line 272: nit: 
```suggestion
	eventBytes, err := protojson.Marshal(event)
```

pkg/server/pubsub_test.go
```diff
@@ -63,23 +66,28 @@ func TestPubSubMessenger_Send(t *testing.T) {
...
-				t.Fatalf("failed to create test PubSub topic: %v", err)
+			eventByte, err := protojson.Marshal(tc.event)
+			if err != nil {
+				t.Errorf("failed to marshal event to byte: %v", err)
 			}
-
-			gotErr := msger.Send(ctx, tc.event)
+			if err != nil {
+				t.Errorf("%v", err)
+			}
```
yolocs, line 84: Duplicate?

pkg/server/pubsub_test.go
```diff
@@ -63,23 +66,28 @@ func TestPubSubMessenger_Send(t *testing.T) {
...
 			if err != nil {
 				t.Fatalf("failed to create new PubSubMessenger: %v", err)
 			}
 
-			// Create the test topic.
-			if _, err := msger.client.CreateTopic(ctx, serverTopicID); err != nil {
-				t.Fatalf("failed to create test PubSub topic: %v", err)
+			eventByte, err := protojson.Marshal(tc.event)
+			if err != nil {
+				t.Errorf("failed to marshal event to byte: %v", err)
```
yolocs, line 80:
```suggestion
				t.Fatalf("failed to marshal event to byte: %v", err)
```

pkg/server/pubsub_test.go
```diff
@@ -63,23 +66,28 @@ func TestPubSubMessenger_Send(t *testing.T) {
...
+			}
+			gotErr := msger.Send(ctx, eventByte, map[string]string{})
 			if diff := testutil.DiffErrString(gotErr, tc.wantErrSubstr); diff != "" {
 				t.Errorf("Process(%+v) got unexpected error substring: %v", tc.name, diff)
 			}
-			if err := msger.Cleanup(); err != nil {
-				t.Errorf("Process(%+v) failed to cleanup: %v", tc.name, err)
-			}
+
+			testTopic.Stop()
```
yolocs, line 90: Should use `t.Cleanup` right after topic creation (in the helper func)

yolocs (changes requested):

pkg/server/event_handler.go
```diff
@@ -270,37 +269,26 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal event to byte: %w", err)
+	}
+
+	attr := map[string]string{}
+
 	if processErr != nil {
 		logger.Errorw(processErr.Error(), "bucketId", m.Attributes["bucketId"], "objectId", m.Attributes["objectId"])
-		if err := h.failureMessenger.Send(ctx, event); err != nil {
+		if err := h.failureMessenger.Send(ctx, eventByte, attr); err != nil {
```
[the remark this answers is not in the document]
sailorlqh: This, the processErr will be in the attr, but this PR is just for refactoring, processErr will be added to attr in the next PR.

pkg/server/pubsub.go
```diff
@@ -19,51 +19,27 @@ import (
...
-	topic := client.Topic(topicID)
-
-	return &PubSubMessenger{client: client, topic: topic}, nil
+func NewPubSubMessenger(topic *pubsub.Topic) *PubSubMessenger {
+	return &PubSubMessenger{topic: topic}
 }
 
 // Send sends a pmap event to a Google Cloud PubSub topic.
-func (p *PubSubMessenger) Send(ctx context.Context, event *v1alpha1.PmapEvent) error {
-	eventBytes, err := protojson.Marshal(event)
```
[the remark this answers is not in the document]
sailorlqh: This Send function's purpose is to publish pubsub messages, we want this function to only handle the publish logic, so we move the logic of marshal PmapEvent to []bytes to the function that calls this `send()` function.

pkg/cli/mapping_server.go
```diff
@@ -93,18 +97,26 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client and topic: %w", err)
+	}
 
-	successMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
+	successMessenger := server.NewPubSubMessenger(successTopic)
+
+	failureTopic, err := createMessangerTopic(ctx, c.cfg.ProjectID, c.cfg.FailureTopicID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event messenger: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create failure event pubsub client and topic: %w", err)
```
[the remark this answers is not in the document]
sqin2019: To make the code cleaner, maybe we can define a closes func array, and the closer will iterate the array to close each of them, below is a rough example. Whenever a new client/resource is created, add the close func to the array. 
```
	var closes []func() error
	closer := func() {
		for _, close := range closes {
			if err := close(); err != nil {
				logger.Errorw("failed to close client", "error", err)
			}
		}
	}
	closes = append(closes, client.Close())

```

pkg/cli/mapping_server.go
sailorlqh: this closer function is called by `defer closer()` in [Run()](https://github.com/abcxyz/pmap/blob/main/pkg/cli/policy_server.go#L59), and when it's executed, we would need to check the error, and log the error there or change the Run() to return the error. which I assume it something we don't want to do?

pkg/cli/mapping_server.go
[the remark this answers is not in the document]
yolocs: not all closer has the same signature. probably introduce more code this way

pkg/cli/mapping_server.go
yolocs: oh I misread `retErr` is defined in the closer. Why do we need it then?


```go
if err := assetClient.Close(); err != nil {
  logger.Errorw("failed to close asset inventory client: %w", err)
}
```

pkg/cli/policy_server.go
```diff
@@ -88,20 +88,20 @@ func (c *PolicyServerCommand) RunUnstarted(ctx context.Context, args []string) (
...
 	handler, err := server.NewHandler(ctx, []server.Processor[*structpb.Struct]{}, successMessenger)
 	if err != nil {
 		return nil, nil, closer, fmt.Errorf("server.NewHandler: %w", err)
 	}
 
 	closer = func() {
-		if err := handler.Cleanup(); err != nil {
-			logger.Errorw("failed to close clean up handler", "error", err)
-		}
+		successTopic.Stop()
```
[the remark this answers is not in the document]
yolocs: To your surprise, you don't have to - see https://pkg.go.dev/cloud.google.com/go/pubsub#Client.Close. But I don't mind we continue to close it since it's trivial to do that

> If the client is available for the lifetime of the program, then Close need not be called at exit.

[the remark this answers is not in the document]
sailorlqh: @sqin2019 and I discussed this, from this reference [https://stackoverflow.com/questions/58351084/does-go-automatically-close-resources-if-not-explicitly-closed], we think it's best to call Close() anywhere to prevent some magic from happening.

pkg/cli/mapping_server.go
sailorlqh: I didn't find this line of code in the current code, can you provide a link to it?

pkg/cli/policy_server.go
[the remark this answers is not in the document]
yolocs: I don't think the reference means anything... If you don't call a func, Go won't call the func. That what it says. But the pubsub doc says you don't have to call it.

pkg/cli/mapping_server.go
```diff
@@ -117,15 +129,31 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+		return nil, nil, closer, fmt.Errorf("failed to create serving infrastructure: %w", err)
+	}
+
 	closer = func() {
-		if err := handler.Cleanup(); err != nil {
-			logger.Errorw("failed to close clean up handler", "error", err)
+		var retErr error
+		successTopic.Stop()
+		failureTopic.Stop()
+		if err := processorClient.Close(); err != nil {
```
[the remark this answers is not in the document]
sailorlqh: They are closed now in closer()

pkg/cli/mapping_server.go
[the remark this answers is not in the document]
sqin2019: true...

pkg/cli/mapping_server.go
```diff
@@ -94,19 +98,27 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
 		return nil, nil, closer, fmt.Errorf("missing PMAP_FAILURE_TOPIC_ID in config")
 	}
 
-	successMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
+	pubsubClient, err := createPubsubClient(ctx, c.cfg.ProjectID)
```
yolocs, line 101: No need to have `createPubsubClient`. What are we trying to achieve with that?

pkg/cli/mapping_server.go
```diff
@@ -94,19 +98,27 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event messenger: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create success event pubsub client: %w", err)
```
yolocs, line 103:
```suggestion
		return nil, nil, closer, fmt.Errorf("failed to create pubsub client: %w", err)
```

pkg/cli/mapping_server.go
```diff
@@ -94,19 +98,27 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
 	}
 
-	failureMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.FailureTopicID)
+	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
+
+	successMessenger := server.NewPubSubMessenger(successTopic)
+
+	failureTopic := pubsubClient.Topic(c.cfg.FailureTopicID)
+
+	failureMessenger := server.NewPubSubMessenger(failureTopic)
```
yolocs, line 112:
```suggestion
	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
	successMessenger := server.NewPubSubMessenger(successTopic)
	failureTopic := pubsubClient.Topic(c.cfg.FailureTopicID)
	failureMessenger := server.NewPubSubMessenger(failureTopic)
```

pkg/cli/policy_server.go
```diff
@@ -88,19 +88,25 @@ func (c *PolicyServerCommand) RunUnstarted(ctx context.Context, args []string) (
...
-	successMessenger, err := server.NewPubSubMessenger(ctx, c.cfg.ProjectID, c.cfg.SuccessTopicID)
+	pubsubClient, err := createPubsubClient(ctx, c.cfg.ProjectID)
 	if err != nil {
-		return nil, nil, closer, fmt.Errorf("failed to create success event messenger: %w", err)
+		return nil, nil, closer, fmt.Errorf("failed to create pubsub client: %w", err)
 	}
 
+	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
+
+	successMessenger := server.NewPubSubMessenger(successTopic)
```
yolocs, line 98:
```suggestion
	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
	successMessenger := server.NewPubSubMessenger(successTopic)
```

pkg/server/event_handler.go
```diff
@@ -270,37 +269,29 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
...
+	}
+
+	attr := map[string]string{}
+
 	if processErr != nil {
 		logger.Errorw(processErr.Error(), "bucketId", m.Attributes["bucketId"], "objectId", m.Attributes["objectId"])
-		if err := h.failureMessenger.Send(ctx, event); err != nil {
+		// TODO(#110): Need to add a LimitReader to read from processErr
+		// before adding it to attr and published by pubsub.
+		// https://cloud.google.com/pubsub/quotas#resource_limits.
```
yolocs, line 283: Move to pubsub.go

pkg/server/pubsub_test.go
```diff
@@ -106,3 +104,25 @@ func newTestPubSubGrpcConn(ctx context.Context, t *testing.T, opts ...pstest.Ser
 
 	return conn
 }
+
+func testCreatePubsubTopic(ctx context.Context, t *testing.T, projectID, topicID string, opts ...option.ClientOption) *pubsub.Topic {
+	t.Helper()
+
+	client, err := pubsub.NewClient(ctx, projectID, opts...)
+	if err != nil {
+		t.Fatalf("failed to create new pubsub client: %v", err)
```
yolocs, line 113:
```suggestion
		t.Fatalf("failed to create pubsub client: %v", err)
```

pkg/server/pubsub_test.go
```diff
@@ -106,3 +104,25 @@ func newTestPubSubGrpcConn(ctx context.Context, t *testing.T, opts ...pstest.Ser
...
+	}
+	if _, err := client.CreateTopic(ctx, serverTopicID); err != nil {
+		t.Fatalf("failed to create test PubSub topic: %v", err)
+	}
+	topic := client.Topic(topicID)
+
+	t.Cleanup(func() {
+		topic.Stop()
+		if err := client.Close(); err != nil {
+			t.Fatalf("failed to close pubsub client: %v", err)
```
yolocs, line 123:
```suggestion
			t.Logf("failed to close pubsub client: %v", err)
```

pkg/cli/mapping_server.go
```diff
@@ -117,15 +129,31 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+		failureTopic.Stop()
+
+		if err := pubsubClient.Close(); err != nil {
+			logger.Errorw("failed to close pubsub client", "error", err)
+		}
+
+		if err := assetClient.Close(); err != nil {
+			logger.Errorw("failed to stop asset client", "error", err)
 		}
 	}
```
yolocs, line 148: Move to be before `srv, err := serving.New(c.cfg.Port)`

@sqin2019 - I thought about what we discussed offline. In case, we return err, `Run()` will not call close func. This is ok because we haven't started serving traffic. 

I agree it would be better if we can close things properly when error happens. But let's fix that separately.

yolocs (approved):

sqin2019 (approved):

pkg/cli/mapping_server.go
sqin2019: sg

Issue #118, opened 2023-06-12
Add DLQ pubsub


sqin2019:
If yes, we can add DLQ in e2e terraform module, so that we can forward the error messages after x number of delivery attempts. https://cloud.google.com/pubsub/docs/handling-failures.

Benefits with DLQ: 
 - configure maximum number of delivery attempts
 - error messages can be found in DLQ 

From @capri-xiyue, 
 - currently we rely on [push backoff (maximum back off is 60s)](https://cloud.google.com/pubsub/docs/push#push_backoff) in push subscription for retry strategy,
 - [message retention (default is 7 days)](https://cloud.google.com/pubsub/docs/replay-overview#subscription_message_retention) for stopping retry,
 - [pubsub metrics](https://cloud.google.com/pubsub/docs/monitoring#monitoring_push_subscriptions) for error message monitoring,
 - and [delivery rate](https://cloud.google.com/pubsub/docs/push#delivery_rate) for traffic control

yolocs:
I think we should have DLQ for any pubsub setup. For pmap, it's a matter of when we want to do that.

sqin2019:
todo list:
- add dlq for bigquery pubsub for both ci and e2e.
- add dlq for gcs notification pubsub in e2e module

capri-xiyue:
> * add backoff for retry both gcs and bq

Since backoff retry is not for DLQ PubSub. I will close this issue for now and create new issue for adding backoff for retry both gcs and bq

PR #121, opened 2023-06-14
feat: add dlq for bigquery subscription

sqin2019:
part of #118, 
Tested locally, this will create a DLQ for each bigquery subscription. 
TODO: update ci in infra to use this new module.


sqin2019:
Haven't check what the retry will look like for dlq, will hold until that.

capri-xiyue (approved):

PR #124, opened 2023-06-14
fix: test

capri-xiyue:


sailorlqh (dismissed):

.github/workflows/snapshot-file-copy.yml
```diff
@@ -43,6 +43,7 @@ jobs:
       - name: 'GCS upload files'
         id: 'gcs-upload-file'
         uses: 'google-github-actions/upload-cloud-storage@a5b77a3bf84da1791719585d327e5f90ae5cb53c' # ratchet:google-github-actions/upload-cloud-storage@v1
+        if: '${{ hashFiles(format(''./{0}/*'', inputs.path)) }}'
```
sqin2019, line 46: is it to check if files match pattern ./{0}/* exist? should it be '${{ hashFiles(format(''./{0}/*'', inputs.path)) != '''' }}'?

yolocs, line 46: Add a comment explain why we need this condition

capri-xiyue: Yes. I think this is a abbreviation of it.

capri-xiyue: done

sqin2019 (approved):

PR #127, opened 2023-06-14
Fix: update terraform ref

sailorlqh:


sailorlqh:
Close this since #126 is faster and already got approved.

PR #129, opened 2023-06-15
Fix: define user facing errors

sailorlqh:
fix: #122

### DIFF
--- a/pkg/mapping/processors/assetinventory.go
+++ b/pkg/mapping/processors/assetinventory.go
@@ -25,6 +25,7 @@ import (
 	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/protoutil"
 	"github.com/abcxyz/pmap/apis/v1alpha1"
+	"github.com/abcxyz/pmap/pkg/pmaperrors"
 	"google.golang.org/api/iterator"
 	"google.golang.org/protobuf/types/known/structpb"
 
@@ -119,7 +120,7 @@ func (p *AssetInventoryProcessor) validateAndEnrich(ctx context.Context, resourc
 	}
 	resource, err := p.getSingleResource(ctx, resourceSearchReq)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get single matched resource: %w", err)
+		return nil, pmaperrors.New(fmt.Sprintf("failed to get single matched resource: %v", err))
 	}
 
 	var ancestors []string
--- a/pkg/mapping/processors/assetinventory.go
+++ b/pkg/mapping/processors/assetinventory.go
@@ -73,8 +73,6 @@ func NewAssetInventoryProcessor(ctx context.Context, client *asset.Client, defau
 func (p *AssetInventoryProcessor) Process(ctx context.Context, resourceMapping *v1alpha1.ResourceMapping) error {
 	logger := logging.FromContext(ctx)
 
-	// TODO(#122): define which errors are user facing errors, and wrap them as pmaperrors.processError
-
 	if resourceMapping.GetResource().GetProvider() != gcpProvider {
 		// Skip non-GCP ResourceMapping
 		logger.Debug("%T: skipping unsupported resource provider %q, want %q", p,
--- a/pkg/mapping/processors/assetinventory.go
+++ b/pkg/mapping/processors/assetinventory.go
@@ -84,7 +84,7 @@ func (p *AssetInventoryProcessor) Process(ctx context.Context, resourceMapping *
 
 	resourceScope, err := parseProject(resourceName)
 	if err != nil {
-		return err
+		return pmaperrors.New(fmt.Sprintf("failed to parse project: %v", err))
 	}
 	// Need defaultResourceScope because resources such as GCS bucket won't include Project info in its resource name.
 	// See details: https://cloud.google.com/asset-inventory/docs/resource-name-format.
@@ -250,7 +250,7 @@ func parseProject(resourceName string) (string, error) {
 		if e == "projects" {
 			if i+1 >= len(s) || s[i+1] == "" {
 				// This is obviously an invalid input.
-				return "", pmaperrors.New(fmt.Sprintf("invalid resource name: %s", resourceName))
+				return "", fmt.Errorf("invalid resource name: %s", resourceName)
 			}
 			project = s[i+1]
 		}
--- a/pkg/mapping/processors/assetinventory.go
+++ b/pkg/mapping/processors/assetinventory.go
@@ -250,7 +250,7 @@ func parseProject(resourceName string) (string, error) {
 		if e == "projects" {
 			if i+1 >= len(s) || s[i+1] == "" {
 				// This is obviously an invalid input.
-				return "", fmt.Errorf("invalid resource name: %s", resourceName)
+				return "", pmaperrors.New(fmt.Sprintf("invalid resource name: %s", resourceName))
 			}
 			project = s[i+1]
 		}
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -138,27 +138,27 @@ func TestMappingEventHandling(t *testing.T) {
 		},
 		// TODO(#122): un-comment the below test once user facing errors are defined.
 		// #122 is blocking this test case from succeeding
-		// {
-		// 	name:          "mapping_failure_event",
-		// 	resourceName:  fmt.Sprintf("//pubsub.googleapis.com/projects/%s/topics/%s", cfg.ProjectID, "non_existent_topic"),
-		// 	bigqueryTable: cfg.MappingFailureTableID,
-		// 	wantResourceMapping: &v1alpha1.ResourceMapping{
-		// 		Resource: &v1alpha1.Resource{
-		// 			Provider: "gcp",
-		// 			Name:     fmt.Sprintf("//pubsub.googleapis.com/projects/%s/topics/%s", cfg.ProjectID, "non_existent_topic"),
-		// 		},
-		// 		Contacts: &v1alpha1.Contacts{Email: []string{"group@example.com"}},
-		// 	},
-		// 	wantGithubSource: &v1alpha1.GitHubSource{
-		// 		RepoName:                   testGithubRepoValue,
-		// 		Commit:                     testGithubCommitValue,
-		// 		Workflow:                   testWorkflowValue,
-		// 		WorkflowSha:                testWorkflowShaValue,
-		// 		WorkflowTriggeredTimestamp: testParseTime(t, testWorkflowTriggeredTimeValue),
-		// 		WorkflowRunId:              testWorkflowRunID,
-		// 		WorkflowRunAttempt:         1,
-		// 	},
-		// },
+		{
+			name:          "mapping_failure_event",
+			resourceName:  fmt.Sprintf("//pubsub.googleapis.com/projects/%s/topics/%s", cfg.ProjectID, "non_existent_topic"),
+			bigqueryTable: cfg.MappingFailureTableID,
+			wantResourceMapping: &v1alpha1.ResourceMapping{
+				Resource: &v1alpha1.Resource{
+					Provider: "gcp",
+					Name:     fmt.Sprintf("//pubsub.googleapis.com/projects/%s/topics/%s", cfg.ProjectID, "non_existent_topic"),
+				},
+				Contacts: &v1alpha1.Contacts{Email: []string{"group@example.com"}},
+			},
+			wantGithubSource: &v1alpha1.GitHubSource{
+				RepoName:                   testGithubRepoValue,
+				Commit:                     testGithubCommitValue,
+				Workflow:                   testWorkflowValue,
+				WorkflowSha:                testWorkflowShaValue,
+				WorkflowTriggeredTimestamp: testParseTime(t, testWorkflowTriggeredTimeValue),
+				WorkflowRunId:              testWorkflowRunID,
+				WorkflowRunAttempt:         1,
+			},
+		},
 	}
 
 	for _, tc := range cases {
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -61,6 +61,12 @@ var (
 	gcsClient *storage.Client
 )
 
+type QueryResult struct {
+	Data       string
+	ProcessErr string
+	PmapEvent  *v1alpha1.PmapEvent
+}
+
 func TestMain(m *testing.M) {
 	os.Exit(func() int {
 		ctx := context.Background()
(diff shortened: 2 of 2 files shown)


test/integration/main_test.go
```diff
@@ -195,12 +201,12 @@ contacts:
 			}
 
 			// Check if the file uploaded exists in BigQuery.
-			queryString := fmt.Sprintf("SELECT data FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
+			queryString := fmt.Sprintf("SELECT data, CASE WHEN STRING(attributes.ProcessErr) is null THEN '' ELSE STRING(attributes.ProcessErr) END as ProcessErr FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
```
capri-xiyue, line 204: Instead of select multiple fields, can you just select the whole bq row and compare it. Selecting multiple fields especially using CASE xxx makes the query quite difficult to understand

test/integration/main_test.go
```diff
@@ -334,11 +343,11 @@ deletion_timeline:
 
 // testGetFirstPmapEventWithRetries queries the DB and get the matched PmapEvent.
 // If no matched PmapEvent was found, the query will be retried with the specified retry inputs.
-func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) *v1alpha1.PmapEvent {
+func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) (*v1alpha1.PmapEvent, string) {
 	tb.Helper()
 
 	b := retry.NewConstant(cfg.QueryRetryWaitDuration)
-	var pmapEvents []*v1alpha1.PmapEvent
+	var logEntries []*QueryResult
```
capri-xiyue, line 350: This is not logentry.

test/integration/main_test.go
```diff
@@ -61,6 +61,12 @@ var (
 	gcsClient *storage.Client
 )
 
+type QueryResult struct {
```
capri-xiyue, line 64: Why the query result looks like this? It looks weird as it doesn't match the bq schema we use.

test/integration/main_test.go
```diff
@@ -348,19 +357,19 @@ func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuer
 
 		// Early exit retry if queried pmap event already found.
 		if len(results) > 0 {
-			pmapEvents = results
+			logEntries = results
```
capri-xiyue, line 360: same here

test/integration/main_test.go
```diff
@@ -348,19 +357,19 @@ func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuer
...
 	}); err != nil {
 		tb.Fatalf("Retry failed: %v.", err)
 	}
-	return pmapEvents[0]
+	return logEntries[0].PmapEvent, logEntries[0].ProcessErr
 }
 
 // queryPmapEvents queries the BQ and checks if pmap events matched the query exists or not and return the results.
-func queryPmapEvents(ctx context.Context, query *bigquery.Query) ([]*v1alpha1.PmapEvent, error) {
+func queryPmapEvents(ctx context.Context, query *bigquery.Query) ([]*QueryResult, error) {
```
capri-xiyue, line 372: The change of such function looks weird to me. I can't understand it. Can you explain?

capri-xiyue (changes requested):

test/integration/main_test.go
sqin2019: +1, it is confusing.

pkg/mapping/processors/assetinventory.go
```diff
@@ -119,7 +118,7 @@ func (p *AssetInventoryProcessor) validateAndEnrich(ctx context.Context, resourc
 	}
 	resource, err := p.getSingleResource(ctx, resourceSearchReq)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get single matched resource: %w", err)
+		return nil, pmaperrors.New(fmt.Sprintf("failed to get single matched resource: %v", err))
 	}
```
sqin2019, line 122: https://github.com/abcxyz/pmap/blob/main/pkg/mapping/processors/assetinventory.go#L86 should also be user facing error

test/integration/main_test.go
sqin2019: +1, it will be easier to understand this way. we can compare if each of them are what we expected separately.

test/integration/main_test.go
sailorlqh: Now select the whole row.

test/integration/main_test.go
sailorlqh: renamed to entires

test/integration/main_test.go
sailorlqh: Changed to only `Data` and `Attributes`

sqin2019 (approved):

pkg/mapping/processors/assetinventory.go
```diff
@@ -251,7 +250,7 @@ func parseProject(resourceName string) (string, error) {
 		if e == "projects" {
 			if i+1 >= len(s) || s[i+1] == "" {
 				// This is obviously an invalid input.
-				return "", fmt.Errorf("invalid resource name: %s", resourceName)
+				return "", pmaperrors.New(fmt.Sprintf("invalid resource name: %s", resourceName))
```
capri-xiyue, line 253: Instead of using pmaperrors here, I'd suggest writing it in https://github.com/abcxyz/pmap/blob/main/pkg/mapping/processors/assetinventory.go#L87-L88.

test/integration/main_test.go
```diff
@@ -231,6 +241,9 @@ contacts:
 					t.Errorf("iamPolicies is blank in resourcemapping.annotations")
 				}
 			}
+			if !strings.Contains(gotProcessError, tc.wantProcessErrSubStr) {
```
capri-xiyue, line 244: Use https://github.com/abcxyz/pkg/blob/683ea9d2f9bd3dcd89ab76f4a2fedfda17a4d977/testutil/error.go#L25

test/integration/main_test.go
```diff
@@ -334,33 +347,39 @@ deletion_timeline:
 
 // testGetFirstPmapEventWithRetries queries the DB and get the matched PmapEvent.
 // If no matched PmapEvent was found, the query will be retried with the specified retry inputs.
-func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) *v1alpha1.PmapEvent {
+func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) (*v1alpha1.PmapEvent, string) {
```
capri-xiyue, line 350: I'd suggest doing some refactor here. Use one function to get the BQ rows with retries and then write another two function to get pmap event and processError. Currently the signature looks confusing, the naming is `testGetFirstPmapEventWithRetries` and the return is one pmapevent and one string(what does the return string means is confusing here)

test/integration/main_test.go
```diff
@@ -61,6 +62,15 @@ var (
 	gcsClient *storage.Client
 )
 
+type BQRow struct {
```
capri-xiyue, line 65: bqEntry. This is only used in test, right? Why use capital case here?

test/integration/main_test.go
```diff
@@ -375,27 +394,20 @@ func queryPmapEvents(ctx context.Context, query *bigquery.Query) ([]*v1alpha1.Pm
 	if err != nil {
 		return nil, retry.RetryableError(fmt.Errorf("failed to read job: %w", err))
 	}
-	var pmapEvents []*v1alpha1.PmapEvent
+
+	var entries []*BQRow
 	for {
-		var row []bigquery.Value
+		var row BQRow
```
capri-xiyue, line 400: Why remove []bigquery.Value here? What error will it throw if BQ can't unmarshal the bq row to the BQRow type you defined?

test/integration/main_test.go
```diff
@@ -334,33 +347,39 @@ deletion_timeline:
...
+		results, err := querysBQRows(ctx, bqQuery)
 		if err != nil {
 			tb.Logf("failed to query pmap events: %v", err)
 			return err
 		}
 
 		// Early exit retry if queried pmap event already found.
 		if len(results) > 0 {
-			pmapEvents = results
+			if err := protojson.Unmarshal([]byte(results[0].Data), &pmapEvent); err != nil {
```
capri-xiyue, line 365: I'd suggest moving these two unmarhsal out of this function and write it outside

capri-xiyue (changes requested):

test/integration/main_test.go
```diff
@@ -334,33 +347,39 @@ deletion_timeline:
...
 		tb.Fatalf("Retry failed: %v.", err)
 	}
-	return pmapEvents[0]
+	return &pmapEvent, attributes.ProcessErr
 }
 
-// queryPmapEvents queries the BQ and checks if pmap events matched the query exists or not and return the results.
-func queryPmapEvents(ctx context.Context, query *bigquery.Query) ([]*v1alpha1.PmapEvent, error) {
+// querysBQRows queries the BQ and checks if a bigqury row matched the query exists or not and return the results.
+func querysBQRows(ctx context.Context, query *bigquery.Query) ([]*BQRow, error) {
```
capri-xiyue, line 382: Make it consistent cross the test code regarding `row` and `entry`

test/integration/main_test.go
```diff
@@ -61,6 +62,15 @@ var (
 	gcsClient *storage.Client
 )
 
+type BQRow struct {
+	Data       string
+	Attributes string
+}
+
+type BQAttributes struct {
```
capri-xiyue, line 70: Same here.

test/integration/main_test.go
```diff
@@ -61,6 +62,15 @@ var (
...
+	ProcessErr string
```
capri-xiyue, line 71: Same here.

test/integration/main_test.go
```diff
@@ -61,6 +62,15 @@ var (
 	gcsClient *storage.Client
 )
 
+type BQRow struct {
+	Data       string
+	Attributes string
```
capri-xiyue, line 67: Does it work that instead of two string here, the type of Data is `pmapevent` and the type of Attributes is the  BQAttributes you defined?

test/integration/main_test.go
sailorlqh: I think this are two strings, why do we want to wrap it two Error and compare?

capri-xiyue: I thought the first one is Error. Then no need to change it.

test/integration/main_test.go
sailorlqh: renamed to `testGetFirstBQEntryWithRetires`, and now this returns a `bqEntry`

pkg/mapping/processors/assetinventory.go
```diff
@@ -85,7 +84,7 @@ func (p *AssetInventoryProcessor) Process(ctx context.Context, resourceMapping *
 
 	resourceScope, err := parseProject(resourceName)
 	if err != nil {
-		return err
+		return pmaperrors.Wrap(err)
```
capri-xiyue, line 87: Why in here use `wrap` but in line 121, using `new`?

test/integration/main_test.go
```diff
@@ -195,12 +210,14 @@ contacts:
...
-			queryString := fmt.Sprintf("SELECT data FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
+			queryString := fmt.Sprintf("SELECT * FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
 			queryString += ` WHERE JSON_VALUE(data.payload.annotations.traceID) = ?`
 			bqQuery := bqClient.Query(queryString)
 			bqQuery.Parameters = []bigquery.QueryParameter{{Value: traceID.String()}}
 
-			gotPmapEvent := testGetFirstPmapEventWithRetries(ctx, t, bqQuery, cfg)
+			gotBQEntry := testGetFirstBQEntryWithRetries(ctx, t, bqQuery, cfg)
+			gotPmapEvent := gotBQEntry.Data
+			gotProcessErrorStr := gotBQEntry.Attributes.ProcessErr
```
capri-xiyue, line 220: No need to write it here, just inline

test/integration/main_test.go
```diff
@@ -334,33 +355,33 @@ deletion_timeline:
 
 // testGetFirstPmapEventWithRetries queries the DB and get the matched PmapEvent.
```
capri-xiyue, line 356: Update comments

test/integration/main_test.go
```diff
@@ -334,33 +355,33 @@ deletion_timeline:
...
 // If no matched PmapEvent was found, the query will be retried with the specified retry inputs.
-func testGetFirstPmapEventWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) *v1alpha1.PmapEvent {
+func testGetFirstBQEntryWithRetries(ctx context.Context, tb testing.TB, bqQuery *bigquery.Query, cfg *config) *bqEntry {
```
capri-xiyue, line 358: testGetFirstMatchedBQEntryWithRetries

pkg/mapping/processors/assetinventory.go
sailorlqh: Changed to use `new` in both cases.

test/integration/main_test.go
sailorlqh: Done.

test/integration/main_test.go
sailorlqh: new changed to `queryBQEntries` and returns BQ entires.

test/integration/main_test.go
```diff
@@ -375,27 +396,41 @@ func queryPmapEvents(ctx context.Context, query *bigquery.Query) ([]*v1alpha1.Pm
 	if err != nil {
 		return nil, retry.RetryableError(fmt.Errorf("failed to read job: %w", err))
 	}
-	var pmapEvents []*v1alpha1.PmapEvent
+
+	var entries []*bqEntry
 	for {
-		var row []bigquery.Value
-		err := it.Next(&row)
+		var entry []bigquery.Value
```
capri-xiyue, line 402: Just curious, can BQ automatically handle the unmarshal when we use `var entry bqEntry` here?

sailorlqh: I tried, it can't, that's why I did it this way.

test/integration/main_test.go
```diff
@@ -52,6 +53,11 @@ const (
 	testWorkflowRunAttempt         = "1"
 )
 
+const (
```
capri-xiyue, line 56: Discussed offline. Using the index is error prone and is difficult to understand. Need to get rid of it.

test/integration/main_test.go
```diff
@@ -332,35 +358,35 @@ deletion_timeline:
 	}
 }
 
-// testGetFirstPmapEventWithRetries queries the DB and get the matched PmapEvent.
+// testGetFirstMatchedBQEntryWithRetries queries the DB and get the matched PmapEvent.
```
capri-xiyue, line 361: `queries the DB and get the matched PmapEvent`, update it. No more pmapevent here.

test/integration/main_test.go
```diff
@@ -332,35 +358,35 @@ deletion_timeline:
...
 // If no matched PmapEvent was found, the query will be retried with the specified retry inputs.
```
capri-xiyue, line 362: Same

capri-xiyue (dismissed):

capri-xiyue (approved):

PR #130, opened 2023-06-20
feat: add release workflow

sailorlqh:
Add docker image release and github release to `release.yml` as a github workflow.

Both release are tested using my test project with some modified parameters like serviceAccount and registry repo. Here are  links to the release in the test project:
- [Artifact registry repo link](https://pantheon.corp.google.com/artifacts/docker/tycho-qinhang-jvs-dev-0a9a/us-central1/pmap-dev/pmap?e=-13802955&jsmode=o&mods=logs_tg_staging&project=tycho-qinhang-jvs-dev-0a9a).

- [Test github repo link](https://github.com/sailorlqh/github-action-test). 

You should be able to find the invite to the above github repo in another email.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,100 @@
+name: 'release'
+on:
+  push:
+    tags:
+      - 'v*'
+
+env:
+  # .goreleaser.docker.yaml reads DOCKER_REPO
+  DOCKER_REPO: 'us-docker.pkg.dev/abcxyz-artifacts/docker-images/pmap'
+  WIF_PROVIDER: 'projects/921163060412/locations/global/workloadIdentityPools/github-pool-3b39/providers/github-provider'
+  WIF_SERVICE_ACCOUNT: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
+  DEV_PROJECT_ID: 'pmap-dev'
+  DEV_REGION: 'us-central1'
+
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+  WIF_SERVICE_ACCOUNT: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
+  DEV_PROJECT_ID: 'pmap-dev'
+  DEV_REGION: 'us-central1'
+
+# Don't cancel in progress since we don't want to have half-baked release.
+concurrency: '${{ github.workflow }}-${{ github.head_ref || github.ref }}-release'
+
+jobs:
+  ci:
+    uses: 'abcxyz/pmap/.github/workflows/ci.yml@main' # ratchet:exclude
```
sethvargo, line 20: I'm not sure we want this? Don't we want "needs" instead?

sailorlqh: JVS has the same pattern here: https://github.com/abcxyz/jvs/blob/main/.github/workflows/release.yml#L17-L19.

The purpose here is to run CI before starting the release process. The line for `needs: [ci]` is a few more lines below at [here](https://github.com/abcxyz/pmap/blob/sailorlqh/add_release_wf/.github/workflows/release.yml#L24).

sethvargo (approved):

capri-xiyue (approved):

sailorlqh:
fix: #130

Issue #103, opened 2023-05-23
Add log info if value is not found.


sailorlqh:
Context: https://github.com/abcxyz/pmap/blob/main/pkg/server/event_handler.go#L342-L364

qiaoweiz:
> Context: https://github.com/abcxyz/pmap/blob/main/pkg/server/event_handler.go#L342-L364

Hi @sailorlqh , I want to double check with the requirement. 
do you only want to log when the value is not found? 
or to log error and return nil (follow the same path as https://screenshot.googleplex.com/JYW5jdoDpqD3HwZ)?

PR #135, opened 2023-06-26
[#103] Add log info if value is not found

qiaoweiz:
Add log info if value for MetadataKeyGitHubCommit, MetadataKeyGitHubRepo, MetadataKeyWorkflow, MetadataKeyWorkflowSha is not found. 

Link to issue: https://github.com/abcxyz/pmap/issues/103


pkg/server/event_handler.go
```diff
@@ -348,19 +348,35 @@ func parseGitHubSource(ctx context.Context, data []byte, objAttrs map[string]str
 
 	var r v1alpha1.GitHubSource
 
-	if c, found := pm.Metadata[MetadataKeyGitHubCommit]; found {
+	// Set github-commit. Response with error message when github-commit can not be found.
+	c, found := pm.Metadata[MetadataKeyGitHubCommit]
+	if !found {
+		return nil, fmt.Errorf("github-commit not found")
```
capri-xiyue, line 354: I think we don't have to return error here. The source may not be from github. Just adding logging info should be enough.

sailorlqh: Agreed, these fields can be empty, we just want a log when these fields are missing.

pkg/server/event_handler.go
```diff
@@ -341,26 +341,44 @@ type notificationPayload struct {
...
 		return nil, fmt.Errorf("failed to unmarshal payloadMetadata %w", err)
 	}
 
 	var r v1alpha1.GitHubSource
 
-	if c, found := pm.Metadata[MetadataKeyGitHubCommit]; found {
+	// Set github-commit.
+	c, found := pm.Metadata[MetadataKeyGitHubCommit]
+	if !found {
+		logger.Infof("github-commit not found")
```
capri-xiyue, line 356: nits: Just use the const defined in the file `MetadataKeyGitHubCommit` instead of repeat the raw string.

pkg/server/event_handler.go
```diff
@@ -341,26 +341,44 @@ type notificationPayload struct {
...
+	} else {
 		r.Commit = c
 	}
 
-	if rn, found := pm.Metadata[MetadataKeyGitHubRepo]; found {
+	// Set github-repo.
+	rn, found := pm.Metadata[MetadataKeyGitHubRepo]
+	if !found {
+		logger.Infof("github-repo not found")
```
capri-xiyue, line 364: Same

pkg/server/event_handler.go
```diff
@@ -341,26 +341,44 @@ type notificationPayload struct {
...
+	} else {
 		r.RepoName = rn
 	}
 
-	if w, found := pm.Metadata[MetadataKeyWorkflow]; found {
+	// Set github-workflow.
+	w, found := pm.Metadata[MetadataKeyWorkflow]
+	if !found {
+		logger.Infof("github-workflow not found")
```
capri-xiyue, line 372: Same

pkg/server/event_handler.go
```diff
@@ -341,26 +341,44 @@ type notificationPayload struct {
...
+	} else {
 		r.Workflow = w
 	}
 
-	if ws, found := pm.Metadata[MetadataKeyWorkflowSha]; found {
+	// Set github-workflow-sha.
+	ws, found := pm.Metadata[MetadataKeyWorkflowSha]
+	if !found {
+		logger.Infof("github-workflow-sha not found")
```
capri-xiyue, line 380: Same

pkg/server/event_handler.go
```diff
@@ -353,31 +353,31 @@ func parseGitHubSource(ctx context.Context, data []byte, objAttrs map[string]str
 	// Set github-commit.
 	c, found := pm.Metadata[MetadataKeyGitHubCommit]
 	if !found {
-		logger.Infof("github-commit not found")
+		logger.Infof("MetadataKeyGitHubCommit not found")
```
capri-xiyue, line 356: Looks like you have a misunderstanding about my previous comment. I mean using the constant here https://github.com/abcxyz/pmap/blob/d001c0d517e11b16284ab36b5fc2d0d0fca0ac81/pkg/server/event_handler.go#L78 . Not using `MetadataKeyGitHubCommit` as a raw string.

pkg/server/event_handler.go
```diff
@@ -353,31 +353,31 @@ func parseGitHubSource(ctx context.Context, data []byte, objAttrs map[string]str
...
 	} else {
 		r.Commit = c
 	}
 
 	// Set github-repo.
 	rn, found := pm.Metadata[MetadataKeyGitHubRepo]
 	if !found {
-		logger.Infof("github-repo not found")
+		logger.Infof("MetadataKeyGitHubRepo not found")
```
capri-xiyue, line 364: Same

pkg/server/event_handler.go
```diff
@@ -353,31 +353,31 @@ func parseGitHubSource(ctx context.Context, data []byte, objAttrs map[string]str
...
 	} else {
 		r.RepoName = rn
 	}
 
 	// Set github-workflow.
 	w, found := pm.Metadata[MetadataKeyWorkflow]
 	if !found {
-		logger.Infof("github-workflow not found")
+		logger.Infof("MetadataKeyWorkflow not found")
```
capri-xiyue, line 372: Same

pkg/server/event_handler.go
```diff
@@ -353,31 +353,31 @@ func parseGitHubSource(ctx context.Context, data []byte, objAttrs map[string]str
...
 	} else {
 		r.Workflow = w
 	}
 
 	// Set github-workflow-sha.
 	ws, found := pm.Metadata[MetadataKeyWorkflowSha]
 	if !found {
-		logger.Infof("github-workflow-sha not found")
+		logger.Infof("MetadataKeyWorkflowSha not found")
```
capri-xiyue, line 380: Same

pkg/server/event_handler_test.go
qiaoweiz, line 1: Tested with "go test" and saw the log 

`"{"severity":"INFO","timestamp":"2023-06-26T23:58:41.226006408Z","caller":"server/event_handler.go:356","message":"MetadataKeyGitHubCommit not found"}"`

pkg/server/event_handler_test.go
```diff
@@ -189,6 +189,33 @@ isOK: true`),
 			wantStatusCode:     http.StatusInternalServerError,
 			wantRespBodySubstr: "failed to unmarshal payloadMetadata",
 		},
+		{
```
capri-xiyue, line 192: The meaning of this test case is not that obvious.

pkg/server/event_handler.go
sailorlqh: something like 
```
logger.Infof("%s not found", MetadataKeyGitHubCommit)
```

pkg/server/event_handler_test.go
qiaoweiz: I don't have strong opinion on it. 
@sailorlqh  What do you think?

sailorlqh: I think it won't hurt to add this test case to make sure the `logger.Infof()` is called without causing any issue.

capri-xiyue: It is not obvious that this test case is to test`logger.Infof()` is called without causing any issue.

sailorlqh: Maybe rename the test case to `success_with_github_commit_missing`.

pkg/server/event_handler.go
qiaoweiz: Updated & marked as resolved.

pkg/server/event_handler.go
qiaoweiz: Updated & marked as resolved.

pkg/server/event_handler.go
qiaoweiz: Updated & marked as resolved.

pkg/server/event_handler.go
qiaoweiz: Updated & marked as resolved.

pkg/server/event_handler.go
qiaoweiz: Updated.

pkg/server/event_handler.go
qiaoweiz: updated.

pkg/server/event_handler.go
qiaoweiz: updated.

pkg/server/event_handler.go
qiaoweiz: updated.

pkg/server/event_handler_test.go
qiaoweiz: updated test's name as `success_with_github_commit_missing`.

pkg/server/event_handler_test.go
qiaoweiz: Updated the log to: 

`{"severity":"INFO","timestamp":"2023-06-27T01:01:47.882440647Z","caller":"server/event_handler.go:356","message":"github-commit not found"}`

sailorlqh (dismissed):
Approved, but let's wait for Xiyue's opinion before merging.

pkg/server/event_handler_test.go
capri-xiyue: Add a comment to explain this test case is to test `logger.Infof() is called without causing any issue.`

qiaoweiz: A comment is added

capri-xiyue (approved):

Issue #128, opened 2023-06-15
validate workflow result in integ test.


sailorlqh:
Call a reuseable (copy/change) workflow, and verify the result in integ test.

PR #136, opened 2023-06-27
feat: test reusable workflow in e2e test

sailorlqh:
fix: #128

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,8 @@ env:
   DOCKER_TAG: '${{ github.sha }}'
 
   TF_MODULE_DIR: './terraform/modules/pmap-service'
-  MAPPING_TF_MODULE_DIR: '${RUNNER_TEMP}/mapping_tf_module'
-  POLICY_TF_MODULE_DIR: '${RUNNER_TEMP}/policy_tf_module'
+  MAPPING_TF_MODULE_DIR: 'mapping_tf_module'
+  POLICY_TF_MODULE_DIR: 'policy_tf_module'
   INTEG_TEST_PROJECT_ID: 'pmap-ci'
   INTEG_TEST_BUCKET_ID: 'pmap-ci'
   INTEG_TEST_BIGQUERY_DATASET_ID: 'pmap_86d8'
@@ -149,10 +149,10 @@ jobs:
           echo "INTEG_TEST_OBJECT_PREFIX_OUTPUT=${INTEG_TEST_OBJECT_PREFIX}" >> "$GITHUB_OUTPUT"
       - name: 'Write mapping tfvars file'
         run: |-
-          mkdir -p '${{ env.MAPPING_TF_MODULE_DIR }}'
-          VAR_FILE='${{ env.MAPPING_TF_MODULE_DIR }}/mapping_ci.tfvars'
+          mkdir -p '${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}'
+          VAR_FILE='${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}/mapping_ci.tfvars'
 
-          touch '${{ env.MAPPING_TF_MODULE_DIR }}/mapping_ci.tfvars'
+          touch ${VAR_FILE}
           echo project_id=\"${{ env.INTEG_TEST_PROJECT_ID }}\" >> ${VAR_FILE};
           echo service_name=\"mapping\" >> ${VAR_FILE};
           echo pmap_container_image=\"${{ env.DOCKER_REPO }}/pmap:${{ env.DOCKER_TAG }}-amd64\" >> ${VAR_FILE};
@@ -168,10 +168,10 @@ jobs:
           echo "MAPPING_TF_VAR_FILE=${VAR_FILE}" >> $GITHUB_ENV;
       - name: 'Write policy tfvars file'
         run: |-
-          mkdir -p '${{ env.POLICY_TF_MODULE_DIR }}'
-          VAR_FILE='${{ env.POLICY_TF_MODULE_DIR }}/policy_ci.tfvars'
+          mkdir -p '${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}'
+          VAR_FILE='${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/policy_ci.tfvars'
 
-          touch '${{ env.POLICY_TF_MODULE_DIR }}/policy_ci.tfvars'
+          touch ${VAR_FILE}
           echo project_id=\"${{ env.INTEG_TEST_PROJECT_ID }}\" >> ${VAR_FILE};
           echo service_name=\"policy\" >> ${VAR_FILE};
           echo pmap_container_image=\"${{ env.DOCKER_REPO }}/pmap:${{ env.DOCKER_TAG }}-amd64\" >> ${VAR_FILE};
@@ -186,20 +186,20 @@ jobs:
           echo "POLICY_TF_VAR_FILE=${VAR_FILE}" >> $GITHUB_ENV;
       - name: 'Init the terraform infrastructure for mapping'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ env.MAPPING_TF_MODULE_DIR }}
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR }} init
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}
+          terraform -chdir=${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }} init
       - name: 'Apply the infrastructure updates for mapping'
         id: 'tf_apply_mapping'
         run: |-
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.MAPPING_TF_VAR_FILE }};
+          terraform -chdir=${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.MAPPING_TF_VAR_FILE }};
       - name: 'Init the terraform infrastructure for policy'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ env.POLICY_TF_MODULE_DIR }}
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} init
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} init
       - name: 'Apply the infrastructure updates for policy'
         id: 'tf_apply_policy'
         run: |-
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.POLICY_TF_VAR_FILE }};
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.POLICY_TF_VAR_FILE }};
       - name: 'Upload terraform files for mapping'
         uses: 'actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05' # actions/upload-artifact@v3
         if: '${{ steps.tf_apply_mapping.outcome == ''success'' }}'
@@ -207,8 +207,8 @@ jobs:
           name: 'terraform-mapping-files-${{ github.run_id}}'
           path: |
             ${{ env.MAPPING_TF_VAR_FILE }}
-            ${{ env.MAPPING_TF_MODULE_DIR }}/terraform.tfstate
-            ${{ env.MAPPING_TF_MODULE_DIR }}/terraform.tfstate.backup
+            ${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}/terraform.tfstate
+            ${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}/terraform.tfstate.backup
 
           retention-days: 1
       - name: 'Upload terraform files for policy'
@@ -218,8 +218,8 @@ jobs:
           name: 'terraform-policy-files-${{ github.run_id}}'
           path: |
             ${{ env.POLICY_TF_VAR_FILE }}
-            ${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate
-            ${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate.backup
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate.backup
           retention-days: 1
 
   # Call reusable workflow
@@ -281,8 +281,8 @@ jobs:
       contents: 'read'
       id-token: 'write'
     env:
-      MAPPING_TF_VAR_FILE: '${RUNNER_TEMP}/mapping_tf_module/mapping_ci.tfvars'
-      POLICY_TF_VAR_FILE: '${RUNNER_TEMP}/policy_tf_module/policy_ci.tfvars'
+      MAPPING_TF_VAR_FILE: 'mapping_tf_module/mapping_ci.tfvars'
+      POLICY_TF_VAR_FILE: 'policy_tf_module/policy_ci.tfvars'
     steps:
       - uses: 'actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c' # ratchet:actions/checkout@v3
       - id: 'auth'
@@ -300,30 +300,30 @@ jobs:
       - name: 'Init the mapping terraform infrastructure for clean up'
         if: '${{ always() }}'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR }} ${{ env.MAPPING_TF_MODULE_DIR }}
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR }} init
+          cp -r ${{ env.TF_MODULE_DIR }} ${{ runner.temp}}/${{ env.MAPPING_TF_MODULE_DIR }}
+          terraform -chdir=${{ runner.temp}}/${{ env.MAPPING_TF_MODULE_DIR }} init
       - name: 'download artifact for mapping'
         if: '${{ always() }}'
         uses: 'actions/download-artifact@e9ef242655d12993efdcda9058dee2db83a2cb9b' # actions/download-artifact@v3
         with:
           name: 'terraform-mapping-files-${{ github.run_id }}'
-          path: '${{ env.MAPPING_TF_MODULE_DIR }}'
+          path: '${{ runner.temp}}/${{ env.MAPPING_TF_MODULE_DIR }}'
       - name: 'Init the policy terraform infrastructure for clean up'
         if: '${{ always() }}'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR }} ${{ env.POLICY_TF_MODULE_DIR }}
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} init
+          cp -r ${{ env.TF_MODULE_DIR }} ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} init
       - name: 'download artifact for policy'
         if: '${{ always() }}'
         uses: 'actions/download-artifact@e9ef242655d12993efdcda9058dee2db83a2cb9b' # actions/download-artifact@v3
         with:
           name: 'terraform-policy-files-${{ github.run_id }}'
-          path: '${{ env.POLICY_TF_MODULE_DIR }}'
+          path: '${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}'
       - name: 'Cleanup the infrastructure for mapping'
         if: '${{ always() }}'
         run: |-
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR }} apply -destroy -auto-approve -var-file=${{ env.MAPPING_TF_VAR_FILE }};
+          terraform -chdir=${{ runner.temp}}/${{ env.MAPPING_TF_MODULE_DIR }} apply -destroy -auto-approve -var-file=${{ runner.temp }}/${{ env.MAPPING_TF_VAR_FILE }};
       - name: 'Cleanup the infrastructure for policy'
         if: '${{ always() }}'
         run: |-
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} apply -destroy -auto-approve -var-file=${{ env.POLICY_TF_VAR_FILE }};
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} apply -destroy -auto-approve -var-file=${{ runner.temp }}/${{ env.POLICY_TF_VAR_FILE }};
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
           version: 'v1.16.1' # Manually pinned
           args: 'release -f .goreleaser.docker.yaml --clean'
 
-  # Deploy integration test infrastructure and run integration tests
+  # Deploy integration test infrastructure
   set_up_integration_infra:
     runs-on: 'ubuntu-latest'
     needs: ['build']
@@ -119,7 +119,7 @@ jobs:
       contents: 'read'
       id-token: 'write'
     outputs:
-      INTEG_TEST_OBJECT_PREFIX_OUTPUT: '${{ steps.output_object_prefix.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}'
+      INTEG_TEST_OBJECT_PREFIX_OUTPUT: '${{ steps.generate_object_prefix.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}'
     steps:
       - uses: 'actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c' # ratchet:actions/checkout@v3
       - id: 'auth'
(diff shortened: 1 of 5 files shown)


sailorlqh:
Added @bradegler  and @verbanicm to get some input on the terraform code as they are the experts.

.github/workflows/ci.yml
```diff
@@ -179,31 +186,159 @@ jobs:
...
+      destination_prefix: 'pmap-ci/policy/${{ needs.set_up_integration_infra.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}'
+      path: 'test/e2e/testdata/fakedata/policy'
+
+  run_integ_test:
+    runs-on: 'ubuntu-latest'
+    needs: ['set_up_integration_infra', 'call_mapping_reusble_workflow', 'call_policy_reusble_workflow']
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    env:
```
sqin2019, line 250: For the commonly used envs, can we set them at the top level?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,159 @@ jobs:
...
+        uses: 'google-github-actions/auth@ef5d53e30bbcd8d0836f4288f5e50ff3e086997d' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ env.WIF_PROVIDER }}'
+          service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
+          token_format: 'access_token'
+      # Write the prefix into GITHUB_ENV which will be used in integration test as a prefix of the
+      # uploaded file's path.
+      - name: 'Write object prefix'
+        run: |
+          echo "INTEG_TEST_OBJECT_PREFIX=${{ needs.set_up_integration_infra.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}" >> $GITHUB_ENV
```
sqin2019, line 274: Why it is written again?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,159 @@ jobs:
...
+        if: '${{ always() }}'
+        run: |-
+          cp -r ${{ env.TF_MODULE_DIR }} ${{ env.MAPPING_TF_MODULE_DIR }}
+          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR }} init
+      - name: 'download artifact for mapping'
+        if: '${{ always() }}'
+        uses: 'actions/download-artifact@v3'
+        with:
+          name: 'terraform-mapping-files'
+          path: '${{ env.MAPPING_TF_MODULE_DIR }}'
```
sqin2019, line 325: How do you ensure it is downloading the right files not the files uploaded by a different workflow run?

sqin2019:
Please correct me if I am wrong, you want to test both reusable workflows (upload file to gcs via reusable workflow) and code logic (manually upload file to gcs) on whatever code changes. It seems redundant. We can test reusable workflow when there are changes to them, if no changes to these workflow files then test the code logic. Therefore, we can have a separate workflow to test reusable workflow. Also smaller PRs are preferred.

.github/workflows/ci.yml
```diff
@@ -140,20 +140,26 @@ jobs:
...
       # CI simultaneously.
       - name: 'Generate object prefix'
+        id: 'generate_object_prefix'
+        run: |
+          echo "INTEG_TEST_OBJECT_PREFIX=${RANDOM}" >> "$GITHUB_ENV"
+      - name: 'Output object prefix'
+        id: 'output_object_prefix'
         run: |
-          echo "INTEG_TEST_OBJECT_PREFIX=$RANDOM" >> $GITHUB_ENV
+          echo "INTEG_TEST_OBJECT_PREFIX_OUTPUT=${INTEG_TEST_OBJECT_PREFIX}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
sailorlqh: This was my first approach, and it didn't work well. So I made them seprate.

.github/workflows/ci.yml
sailorlqh: Because the first time we write this, it was in a different job, when switch to a new job, the `github_env` doesn't exist anymore. And this value is required in the integ test, the config get this value for env, and set it as the path prefix for uploading yaml files.

sqin2019: I wonder if we can set all the env vars at the top level...

.github/workflows/ci.yml
sailorlqh: I will add a random postfix to ensure that.

.github/workflows/ci.yml
sailorlqh: Moved to top.

sailorlqh:
> Please correct me if I am wrong, you want to test both reusable workflows (upload file to gcs via reusable workflow) and code logic (manually upload file to gcs) on whatever code changes. It seems redundant. We can test reusable workflow when there are changes to them, if no changes to these workflow files then test the code logic. Therefore, we can have a separate workflow to test reusable workflow. 

The idea for this update is to ensure the workflow's behavior is correct. According to previous discuss with Chen, we decide to add this to the CI e2e test. 

> Also smaller PRs are preferred.
I don't think I can further break down this PR, as calling a reusable workflow in `ci.yml` will require the related terraform code change, and also the tests should be updated.


> Another way is to test reusable workflow -> only check if files successfully uploaded to gcs in ci, test the entire e2e in dev when doing the release, this way we don't need to change the current ci infrastructure. Only concern is what if e2e fail, we need to fix it asap and release a new version.
To make sure the reusable workflow can upload files to GCS, we still need to call the workflow, thus the terraform code change is inevitable. And we didn't make any ci infra change, we just divided them into 3 different jobs. 
Testing in dev will require us to update the cloud run images if we made modification in the code, and I don't think doing a new release every time we makes modification to workflow files is the optimal way.

The current way we can test all the things in ci project, and the current set up can ensure the env are cleaned up after test is running, these are the reason I implemented this way.

bradegler (changes requested):

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR}} init
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ env.POLICY_TF_MODULE_DIR }}
+          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} init
       - name: 'Apply the infrastructure updates for policy'
         id: 'tf_apply_policy'
         run: |-
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR}} apply -auto-approve -var-file=${{env.POLICY_TF_VAR_FILE}};
+          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.POLICY_TF_VAR_FILE }};
+      - name: 'Upload terraform files for mapping'
+        uses: 'actions/upload-artifact@v3'
```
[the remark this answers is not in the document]
sailorlqh: Done.

.github/workflows/ci.yml
```diff
@@ -17,6 +17,24 @@ env:
   DOCKER_REPO: 'us-docker.pkg.dev/pmap-ci/ci-images'
   DOCKER_TAG: '${{ github.sha }}'
 
+  TF_MODULE_DIR: './terraform/modules/pmap-service'
+  MAPPING_TF_MODULE_DIR: '/tmp/mapping_tf_module'
```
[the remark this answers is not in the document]
sailorlqh: changed to use `${{ runner.temp }}`

bradegler (dismissed):

.github/workflows/ci.yml
```diff
@@ -17,6 +17,26 @@ env:
   DOCKER_REPO: 'us-docker.pkg.dev/pmap-ci/ci-images'
   DOCKER_TAG: '${{ github.sha }}'
 
+  TF_MODULE_DIR: './terraform/modules/pmap-service'
+  MAPPING_TF_MODULE_DIR: 'mapping_tf_module'
+  # MAPPING_TF_VAR_FILE: 'mapping_tf_module/mapping_ci.tfvars'
```
sqin2019, line 22: Remove?

.github/workflows/ci.yml
```diff
@@ -17,6 +17,26 @@ env:
...
+  POLICY_TF_MODULE_DIR: 'policy_tf_module'
+  # POLICY_TF_VAR_FILE: 'policy_tf_module/policy_ci.tfvars'
```
sqin2019, line 24: Same here?

.github/workflows/ci.yml
```diff
@@ -179,31 +188,144 @@ jobs:
...
+  clean_up_integration_infra:
+    runs-on: 'ubuntu-latest'
+    if: '${{ always() }}'
+    needs: ['set_up_integration_infra', 'run_integ_test']
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    env:
+      MAPPING_TF_VAR_FILE: '${RUNNER_TEMP}/mapping_tf_module/mapping_ci.tfvars'
+      POLICY_TF_VAR_FILE: '${RUNNER_TEMP}/policy_tf_module/policy_ci.tfvars'
```
sqin2019, line 287: I don't think these files are ever used, they should be empty as they are in a separate job, runner tmp dir is emptied for each job.
Checking the upload and download, the tfvars file is also uploaded and downloaded. I am guessing it is using the tfvar downloaded instead of the one you gave.
Screenshot 2023-06-30 at 11 12 56 AM
Screenshot 2023-06-30 at 11 13 52 AM

sailorlqh: The `MAPPING_TF_VAR_FILE`, and `POLICY_TF_VAR_FILE` are just the file path, after downloading the artifacts, the files will be in the path, which we use to do destroy.

sqin2019: So ${{ runner.temp }} and ${RUNNER_TEMP} is the same thing? (QQ though, if the var files are in the same terraform dir, do we still need to explicit provide tfvar path?)

sailorlqh: Yes, `${ runner.temp}` is only valid in jobs, can't use in `env` variables. I will update this a bit to get rid of the ${RUNNER_TEMP}.

sqin2019: Thanks, just checked, the file has to be exactly terraform.tfvars or terraform.tfvars.json otherwise need to specified that.

test/integration/main_test.go
```diff
@@ -358,6 +358,149 @@ deletion_timeline:
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx := context.Background()
+
+			t.Logf("using trace ID %s", tc.traceID)
+			t.Logf("using workflow run ID %s", cfg.WorkflowRunID)
+			// Check if the file uploaded exists in BigQuery.
+			queryString := fmt.Sprintf("SELECT * FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
+			queryString += `WHERE JSON_VALUE(data.githubSource.workflowRunId) = ?`
```
sailorlqh, line 480: We use workflowRunId instead of traceID here since in this case traceID is provided by yaml file so it's not unique, while workflowRunId is. And this also checks the GCS object metadata is successfully uploaded and ended up in bigquery. So later we only need to compare payload and traceID.

.github/workflows/ci.yml
```diff
@@ -95,31 +112,14 @@ jobs:
           args: 'release -f .goreleaser.docker.yaml --clean'
 
   # Deploy integration test infrastructure and run integration tests
```
capri-xiyue, line 114: update comments

capri-xiyue, line 114: update comments

.github/workflows/ci.yml
```diff
@@ -140,20 +140,26 @@ jobs:
       # Generate a unique object ID prefix to mark the files in GCS and BigQuery so that we run multiple
       # CI simultaneously.
       - name: 'Generate object prefix'
+        id: 'generate_object_prefix'
         run: |
-          echo "INTEG_TEST_OBJECT_PREFIX=$RANDOM" >> $GITHUB_ENV
+          echo "INTEG_TEST_OBJECT_PREFIX=${RANDOM}" >> "$GITHUB_ENV"
+      - name: 'Output object prefix'
```
capri-xiyue, line 146: Can you combine generate and output object prefix in one step since each step is just one echo?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
       - name: 'Apply the infrastructure updates for mapping'
         id: 'tf_apply_mapping'
         run: |-
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR}} apply -auto-approve -var-file=${{env.MAPPING_TF_VAR_FILE}};
+          terraform -chdir=${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.MAPPING_TF_VAR_FILE }};
       - name: 'Init the terraform infrastructure for policy'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR}} ${{ env.POLICY_TF_MODULE_DIR}}
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR}} init
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}
```
capri-xiyue, line 197: Do you need `/*` here?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
           echo "POLICY_TF_VAR_FILE=${VAR_FILE}" >> $GITHUB_ENV;
       - name: 'Init the terraform infrastructure for mapping'
         run: |-
-          cp -r ${{ env.TF_MODULE_DIR}} ${{ env.MAPPING_TF_MODULE_DIR}}
-          terraform -chdir=${{ env.MAPPING_TF_MODULE_DIR}} init
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.MAPPING_TF_MODULE_DIR }}
```
capri-xiyue, line 189: Same, do you need `/*` here?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+          cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} init
       - name: 'Apply the infrastructure updates for policy'
         id: 'tf_apply_policy'
         run: |-
-          terraform -chdir=${{ env.POLICY_TF_MODULE_DIR}} apply -auto-approve -var-file=${{env.POLICY_TF_VAR_FILE}};
+          terraform -chdir=${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }} apply -auto-approve -var-file=${{ env.POLICY_TF_VAR_FILE }};
+      - name: 'Upload terraform files for mapping'
+        uses: 'actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05' # actions/upload-artifact@v3
+        if: '${{ steps.tf_apply_mapping.outcome == ''success'' }}'
```
capri-xiyue, line 205: If the terraform files has some bugs and it encountered failure after creating some resources, you will end up those resources without cleanup as you don't upload terraform files?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+          path: |
+            ${{ env.POLICY_TF_VAR_FILE }}
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate.backup
+          retention-days: 1
+
+  # Call reusable workflow
+  call_mapping_reusable_workflow:
+    needs: ['set_up_integration_infra']
+    uses: 'abcxyz/pmap/.github/workflows/snapshot-file-copy.yml@main' # ratchet:exclude
```
capri-xiyue, line 228: If you use main here, you can't catch the bugs exist in the PR that introduce bugs.  You need to use the version of the PR that runs the CI/CD workflow.

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+    with:
+      workload_identity_provider: 'projects/921163060412/locations/global/workloadIdentityPools/github-pool-3b39/providers/github-provider'
+      service_account: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
+      destination_prefix: 'pmap-ci/mapping/${{ needs.set_up_integration_infra.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}'
+      path: 'test/e2e/testdata/fakedata/mapping'
+
+  call_policy_reusable_workflow:
+    needs: ['set_up_integration_infra']
+    uses: 'abcxyz/pmap/.github/workflows/snapshot-file-copy.yml@main' # ratchet:exclude
```
capri-xiyue, line 237: Same

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate
+            ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}/terraform.tfstate.backup
+          retention-days: 1
+
+  # Call reusable workflow
+  call_mapping_reusable_workflow:
+    needs: ['set_up_integration_infra']
+    uses: 'abcxyz/pmap/.github/workflows/snapshot-file-copy.yml@main' # ratchet:exclude
+    with:
+      workload_identity_provider: 'projects/921163060412/locations/global/workloadIdentityPools/github-pool-3b39/providers/github-provider'
```
capri-xiyue, line 230: ${{ env.WIF_PROVIDER }} can't be used here?

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+      service_account: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
```
capri-xiyue, line 231: same

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+      destination_prefix: 'pmap-ci/mapping/${{ needs.set_up_integration_infra.outputs.INTEG_TEST_OBJECT_PREFIX_OUTPUT }}'
+      path: 'test/e2e/testdata/fakedata/mapping'
+
+  call_policy_reusable_workflow:
+    needs: ['set_up_integration_infra']
+    uses: 'abcxyz/pmap/.github/workflows/snapshot-file-copy.yml@main' # ratchet:exclude
+    with:
+      workload_identity_provider: 'projects/921163060412/locations/global/workloadIdentityPools/github-pool-3b39/providers/github-provider'
```
capri-xiyue, line 239: same

.github/workflows/ci.yml
```diff
@@ -179,31 +186,144 @@ jobs:
...
+      service_account: 'pmap-3b39-ci-sa@pmap-ci.iam.gserviceaccount.com'
```
capri-xiyue, line 240: same

test/integration/main_test.go
capri-xiyue: Can you add comments?

test/integration/main_test.go
```diff
@@ -358,6 +358,149 @@ deletion_timeline:
...
+
+			bqQuery := bqClient.Query(queryString)
+			bqQuery.Parameters = []bigquery.QueryParameter{{Value: cfg.WorkflowRunID}}
```
capri-xiyue, line 483: Do you need this when you use queryString as `WHERE JSON_VALUE(data.githubSource.workflowRunId) = ?`

test/integration/main_test.go
```diff
@@ -358,6 +358,149 @@ deletion_timeline:
...
+		name          string
+		bigqueryTable string
+		traceID       string
+		wantPayload   *structpb.Struct
+	}{
+		{
+			name:          "policy-copy-success-event",
+			bigqueryTable: cfg.PolicyTableID,
+			traceID:       "fake-pmap-dev-manual-test-policy",
+			wantPayload: &structpb.Struct{
```
capri-xiyue, line 446: Don't use resource mapping example for policy

test/integration/main_test.go
```diff
@@ -358,6 +358,149 @@ deletion_timeline:
...
+			ctx := context.Background()
+
+			t.Logf("using trace ID %s", tc.traceID)
+			t.Logf("using workflow run ID %s", cfg.WorkflowRunID)
+			// Check if the file uploaded exists in BigQuery.
+			queryString := fmt.Sprintf("SELECT * FROM `%s.%s.%s`", cfg.ProjectID, cfg.BigQueryDataSetID, tc.bigqueryTable)
+			queryString += `WHERE JSON_VALUE(data.githubSource.workflowRunId) = ?`
+
+			bqQuery := bqClient.Query(queryString)
+			bqQuery.Parameters = []bigquery.QueryParameter{{Value: cfg.WorkflowRunID}}
```
capri-xiyue, line 401: Do you need this when you use queryString as WHERE JSON_VALUE(data.githubSource.workflowRunId) = ?

test/integration/main_test.go
```diff
@@ -358,6 +358,149 @@ deletion_timeline:
...
+	cases := []struct {
+		name                string
+		bigqueryTable       string
+		traceID             string
+		wantResourceMapping *v1alpha1.ResourceMapping
+	}{
+		{
+			name:          "mapping-copy-success-event",
+			bigqueryTable: cfg.MappingTableID,
+			traceID:       "fake-pmap-dev-manual-test-mapping",
```
capri-xiyue, line 375: Can you just set the value of traceID to the workflow run id and then we can make the query string consistent?

capri-xiyue (changes requested):

.github/workflows/ci.yml
sailorlqh: This was one of Suhong's suggestion, and was my first approach, it was a bit problematic, so I separated it.

.github/workflows/ci.yml
sailorlqh: Becase the `POLICY_TF_MODULE_DIR` is mkdir-ed in a previous step already, so we only want the content in `TF_MODULE_DIE` to be copied. Without `/*` here, after the copy command the file structure would be
```
|- POLICY_TF_MODULE_DIR
   |- xxxx.tfvar
   |- TF_MODULE_DIR
       |- ***.tf
```

.github/workflows/ci.yml
sailorlqh: I don't think we can pass env var to reusable workflows.

test/integration/main_test.go
capri-xiyue: Sorry, took a look at the wrong code. Just ignore this

test/integration/main_test.go
capri-xiyue: In addition, can you just set the value of traceID to the workflow run id and then we can make the query string consistent?

.github/workflows/ci.yml
capri-xiyue: Just curious, what was the problem when you combine them in one step?

sailorlqh: when using `${{needs.xxxx.INTEG_TEST_OBJECT_PREFIX}}` in the following step, the value is empty.

.github/workflows/ci.yml
capri-xiyue: I got a little bit confused. At first, mu question is we used `cp -r ${{ env.TF_MODULE_DIR}} ${{ env.POLICY_TF_MODULE_DIR}}` and you changed it to `cp -r ${{ env.TF_MODULE_DIR }}/* ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}` I understand the change from `${{ env.POLICY_TF_MODULE_DIR}}` to ` ${{ runner.temp }}/${{ env.POLICY_TF_MODULE_DIR }}` as you use ${{ runner.temp }}. Based on my understanding, -r already copies file recursively under ${{ env.TF_MODULE_DIR}}

.github/workflows/ci.yml
sailorlqh: Updated to always()

.github/workflows/ci.yml
sailorlqh: Fixed

test/integration/main_test.go
sailorlqh: Discussed offline, the traceID is from the yaml file, which makes it a lot of extra work to enable this.

.github/workflows/ci.yml
sailorlqh: Discussed offline.

.github/workflows/ci.yml
capri-xiyue: By the way, looks like you are using INTEG_TEST_OBJECT_PREFIX and INTEG_TEST_OBJECT_PREFIX_OUTPUT with same value. Maybe you syntax has something wrong.
Does the following work?
1. define `OBJECT_PREFIX` variable = {RANDOM}
2. echo "INTEG_TEST_OBJECT_PREFIX=${OBJECT_PREFIX}" >> "$GITHUB_ENV"
3. echo "INTEG_TEST_OBJECT_PREFIX_OUTPUT=${OBJECT_PREFIX}" >> "$GITHUB_OUTPUT"

test/integration/main_test.go
```diff
@@ -358,6 +355,155 @@ deletion_timeline:
...
+			cmpOpts := []cmp.Option{
+				protocmp.Transform(),
+				protocmp.IgnoreFields(&v1alpha1.ResourceMapping{}, "annotations"),
+			}
+
+			if diff := cmp.Diff(tc.wantResourceMapping, gotResourceMapping, cmpOpts...); diff != "" {
+				t.Errorf("gotPayload unexpected diff (-want,+got):\n%s", diff)
+			}
+
+			if diff := cmp.Diff(tc.traceID, gotResourceMapping.Annotations.AsMap()["traceID"]); diff != "" {
```
capri-xiyue, line 428: Nits: Do you need to compare the trace id here? :

capri-xiyue (dismissed):

test/integration/main_test.go
sailorlqh: Yes, to test the right file is uploaded.

test/integration/main_test.go
sailorlqh: Updated.

test/integration/main_test.go
sailorlqh: Comment added, discussed the traceID thing offline, and decided not to use it.

.github/workflows/ci.yml
```diff
@@ -139,14 +139,12 @@ jobs:
           terraform_wrapper: false
       # Generate a unique object ID prefix to mark the files in GCS and BigQuery so that we run multiple
       # CI simultaneously.
-      - name: 'Generate object prefix'
+      - name: 'Generate and output object prefix'
         id: 'generate_object_prefix'
```
capri-xiyue, line 143: update the id?

capri-xiyue (approved):

.github/workflows/ci.yml
sailorlqh: Updated the name and kept the id.

PR #138, opened 2023-06-29
fix: Enable user to define resource scope

### LINKED ISSUE #137: Enable asset inventory processor customized scope in Mapping Server
### TL;DR

Currently, https://github.com/abcxyz/pmap/blob/5b24e38b19e0c8b8e12854ead228111b16e503af/pkg/cli/mapping_server.go#L123, mapping server defaults to use the project as default scope of Asset Inventory Processor. Customers may want to use folder id or organization id as the default scope of Asset Inventory Processor（otherwise customers may not be able register resources like gcs bucket whose resource name doesn't include project). We need to enable asset inventory processor customized scope in Mapping Server. Similar to how JVS users can customize the grace period, disable period and so on.
### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

sailorlqh:
fix #137


pkg/server/config.go
```diff
@@ -39,6 +41,13 @@ func (cfg *HandlerConfig) Validate() error {
 		return fmt.Errorf("PMAP_SUCCESS_TOPIC_ID is empty and requires a value")
 	}
 
+	if cfg.ResourceScope != "" &&
+		!strings.HasPrefix(cfg.ResourceScope, "projects/") &&
```
sqin2019, line 45: We should make allowed prefixes a global var.

pkg/cli/mapping_server.go
```diff
@@ -120,7 +120,14 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
 	}
 	closer = multicloser.Append(closer, assetClient.Close)
 
-	processor, err := processors.NewAssetInventoryProcessor(ctx, assetClient, fmt.Sprintf("projects/%s", c.cfg.ProjectID))
+	var defaultScope string
+	if c.cfg.ResourceScope == "" {
```
sqin2019, line 124: Should we change the naming to DefaultResourceScope?

pkg/server/config.go
```diff
@@ -39,6 +41,13 @@ func (cfg *HandlerConfig) Validate() error {
...
+		!strings.HasPrefix(cfg.ResourceScope, "folders/") &&
+		!strings.HasPrefix(cfg.ResourceScope, "organizations/") {
+		return fmt.Errorf("ResourceScope doesn't have a valid value")
```
sqin2019, line 48: Maybe more info in the error message, why it is not valid.

terraform/modules/pmap-service/variables.tf
```diff
@@ -73,3 +73,17 @@ variable "gcs_events_filter" {
   type        = string
   default     = null
 }
+
+variable "pmap_resource_scope" {
+  description = <<EOT
+        "This is the scope for where the resources resides in.
+        Options can be one of the following:
+        projects/{PROJECT_ID}
+        projects/{PROJECT_NUMBER}
```
sqin2019, line 82: What is the difference between project id and number, did you mean project name? Should we accept name? I remember an issue last week was related to the scope was project name.

terraform/modules/pmap-service/variables.tf
```diff
@@ -73,3 +73,17 @@ variable "gcs_events_filter" {
...
+        folders/{FOLDER_NUMBER}
+        organizations/{ORGANIZATION_NUMBER}
```
sqin2019, line 84: To be consistent with projects, ID or Number?

pkg/cli/mapping_server.go
```diff
@@ -120,7 +120,14 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
+		defaultScope = fmt.Sprintf("projects/%s", c.cfg.ProjectID)
```
sqin2019, line 125: This project ID is used when we creating the pubsub client for sucess and failure messengers which is the pmap project. I don't think we should use this as default for searching cloud resource assets, an option is to make DefaultResourceScope required.

sailorlqh: The current behavior was defaulting to scope to the current project using the project ID, I just want to keep that behavior.

terraform/modules/pmap-service/variables.tf
sailorlqh: It's intended to be project/number and project/id. 

You can refer to this: https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/searchAllResources#path-parameters. These are the acceptable parameters allowed by the API for searching resources.

sqin2019: Got it, thanks!

pkg/server/config.go
sailorlqh: I thought about making it a global var in `const()`, but they are only used here, so I though putting it here will make the prefix more obvious.

pkg/cli/mapping_server.go
sailorlqh: In the current implementation, the DefaultResourceScope is the project scope. If we decided to make the `ResourceScope` a required value, I think we can rename this to `DefaultResourceScope`, but if we intend to keep the current behavior, I don't think the renaming is necessary.

sqin2019: This resource scope is only used when we cannot find the resource scope in the resource name, so it is more like a default, naming it ResourceScope is confusing?

pkg/cli/mapping_server.go
sqin2019: That's too implicit, we can make it more explicit by making support GCS configurable, when it is on (support GCS), default resource scope is required. Another easier option is to make default resource scope required for all cases, since our customer can easily use their gcp org as the default scope.

pkg/server/config.go
```diff
@@ -27,8 +28,15 @@ type HandlerConfig struct {
...
 	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
 	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
+	ResourceScope  string `env:"PMAP_RESOURCE_SCOPE"`
 }
 
+const (
+	ProjectResourceScopePrefix      = "projects"
+	FolderResourceScopePrefix       = "folders"
+	OrganizationResourceScopePrefix = "organizations"
+)
```
sqin2019, line 38: A cleaner way is use a map (set in golang)
```
var supportedScope = map[string]struct{}{
		"projects": {},
		"folders": {},
		"organizations": {},
	}
```
We can do checks like:
```
scope := strings.Split(cfg.ResourceScope, "/")[0]
if _, ok := supportedScope[scope]; !ok {
 ....
}
```

pkg/server/config_test.go
```diff
@@ -57,14 +57,24 @@ func TestConfig_Validate(t *testing.T) {
...
+			name: "invalid_resource_scope",
+			cfg: &HandlerConfig{
+				Port:           "8080",
+				ProjectID:      testProjectID,
+				SuccessTopicID: testSuccessTopicID,
+				FailureTopicID: testFailureTopicID,
+				ResourceScope:  "foo/bar",
+			},
+			wantErr: "ResourceScope: foo/bar doesn't have a valid value",
+		},
```
sqin2019, line 70: Need a test with a valid and non-empty resource scope.

pkg/cli/mapping_server.go
sailorlqh: Made resource scope a required value.

terraform/e2e/main.tf
```diff
@@ -56,4 +58,5 @@ module "policy_service" {
   pmap_service_account = module.common_infra.run_service_account
   oidc_service_account = module.common_infra.oidc_service_account
   gcs_events_filter    = var.policy_gcs_events_filter
+  pmap_resource_scope  = local.pmap_resource_scope
```
sqin2019, line 61: Resource scope is not required for policy service.

pkg/server/config.go
```diff
@@ -16,21 +16,30 @@ package server
...
+}
+
+var SupportedResourceScope = map[string]struct{}{
+	"projects":      {},
+	"folders":       {},
+	"organizations": {},
 }
 
 // Validate validates the handler config after load.
 func (cfg *HandlerConfig) Validate() error {
```
sqin2019, line 41: I wonder how policy_server_test.go passed, as we don't have resource scope there....
We will need two validate, one for mapping - ValidateMappingConfig() and one for policy - ValidatePolicyConfig(), The default resource scope is only needed in mapping.

.github/workflows/ci.yml
```diff
@@ -179,6 +180,7 @@ jobs:
           echo pmap_service_account=\"${{ env.INTEG_TEST_RUN_SERVICE_ACCOUNT }}\" >> ${VAR_FILE};
           echo oidc_service_account=\"${{ env.INTEG_TEST_OIDC_SERVICE_ACCOUNT }}\" >> ${VAR_FILE};
           echo gcs_events_filter=\"hasPrefix\(attributes.objectId, \\\"policy/${INTEG_TEST_OBJECT_PREFIX}\\\"\)\">> ${VAR_FILE};
+          echo pmap_resource_scope=\"projects/${{ env.INTEG_TEST_PROJECT_ID }}\" >> ${VAR_FILE}
```
sqin2019, line 183: Policy service does not need resource scope.

pkg/server/config.go
```diff
@@ -16,21 +16,30 @@ package server
...
 type HandlerConfig struct {
-	Port           string `env:"PORT,default=8080"`
-	ProjectID      string `env:"PROJECT_ID,required"`
-	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
-	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
+	Port                 string `env:"PORT,default=8080"`
+	ProjectID            string `env:"PROJECT_ID,required"`
+	SuccessTopicID       string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
+	FailureTopicID       string `env:"PMAP_FAILURE_TOPIC_ID"`
+	DefaultResourceScope string `env:"PMAP_RESOURCE_SCOPE,required"`
```
sqin2019, line 31: Please add comment specifying this field is only required for mapping service, and remove required. since the naming is changed, please change the env name accordingly.

terraform/modules/pmap-service/variables.tf
```diff
@@ -73,3 +73,16 @@ variable "gcs_events_filter" {
...
+        When use default (empty), we will default to projects/{PROJECT_ID}"
```
sqin2019, line 85: This line should be removed?

terraform/e2e/main.tf
sailorlqh: No it's not, but both mapping and policy use the same terraform/pmap-service module.
The resource scope is defined as required in variables.tf, so I have to provide it a value.

pkg/server/config.go
sailorlqh: The tests are passed because there are no `success` test cases. We can add the case as a follow up, as we need fake clients for pub/sub, asset, and server.

pkg/server/config.go
sailorlqh: Added comments, I think we can keep the env name, as it matches `PMAP_FAILURE_TOPIC_ID`, and they are both only used in mapping service.

terraform/e2e/main.tf
sqin2019: Forgot to remove?

terraform/e2e/main.tf
```diff
@@ -15,6 +15,7 @@
 locals {
   mapping_service_name = "mapping"
   policy_service_name  = "policy"
+  pmap_resource_scope  = "projects/pmap-dev"
```
sqin2019, line 18: Overlooked this, we should make this an input variable. Once this change merged, we need to update https://github.com/abcxyz/infra/blob/main/terraform/legacy/pmap/dev/main.tf#L22

pkg/server/config.go
```diff
@@ -26,7 +26,15 @@ type HandlerConfig struct {
 	Port           string `env:"PORT,default=8080"`
 	ProjectID      string `env:"PROJECT_ID,required"`
 	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
-	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
+	// FailureTopicID and DefaultResourceScope are only used in mapping service
+	FailureTopicID       string `env:"PMAP_FAILURE_TOPIC_ID"`
+	DefaultResourceScope string `env:"PMAP_RESOURCE_SCOPE"`
```
sqin2019, line 31: I just came up with a better idea:
```
type HandlerConfig struct {
	Port           string `env:"PORT,default=8080"`
	ProjectID      string `env:"PROJECT_ID,required"`
	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`.  // note we might need failure topic in policy
        // MappingConfig is only required for mapping service.
	MappingConfig MappingConfig
}

type MappingConfig struct {
	DefaultResourceScope string `env:"PMAP_RESOURCE_SCOPE"`
}

// Validate mapping service config.
func (cfg *HandlerConfig) ValidateMappingConfig() error {
	if err := cfg.Validate(); err != nil {
		return fmt.Errorf("...")
	}
	// if cfg.MappingConfig.DefaultResourceScope...
	return nil
}
```

pkg/server/config.go
sqin2019: To correct, PMAP_FAILURE_TOPIC_ID is optional for policy, there is a case when user customize a policyprocessor and want to send the failure event to failure table.

terraform/e2e/main.tf
sailorlqh: Will do it after this. Created #139.

pkg/server/config.go
sailorlqh: Done.

pkg/server/config.go
```diff
@@ -42,6 +57,27 @@ func (cfg *HandlerConfig) Validate() error {
 	return nil
 }
 
+// Validate validates the handler config after load.
+func (cfg *HandlerConfig) ValidateMappingConfig() error {
```
sqin2019, line 61: Unit test?

terraform/e2e/main.tf
sqin2019: This is an e2e terraform we recommend customer to use, better not hardcode anything us specific. It won't break the dev env since it has pined the past commit in infra. We can do it in the same PR, or leave this e2e module untouched, and fix it in the next PR.

pkg/server/config.go
sailorlqh: Added

terraform/e2e/main.tf
sailorlqh: Updated, removed this local var, and made it a variable in variables.tf

pkg/cli/mapping_server.go
```diff
@@ -97,6 +97,10 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
 	}
 	logger.Debugw("loaded configuration", "config", c.cfg)
 
+	if err := c.cfg.ValidateMappingConfig(); err != nil {
+		return nil, nil, closer, fmt.Errorf("invalid mapping configuration: %w", err)
+	}
+
 	// For mapping server, we also require a failure topic ID.
 	if c.cfg.FailureTopicID == "" {
```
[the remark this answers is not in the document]
sailorlqh: Moved FailureTopicID check to `ValidateMappingConfig` function.

terraform/e2e/variables.tf
```diff
@@ -56,3 +56,16 @@ variable "bigquery_table_delete_protection" {
...
+        "This is the scope for where the resources resides in.
+        Options can be one of the following:
+        projects/{PROJECT_ID}
+        projects/{PROJECT_NUMBER}
+        folders/{FOLDER_NUMBER}
+        organizations/{ORGANIZATION_NUMBER}
+    EOT
+  type        = string
+  default     = ""
+}
```
sqin2019, line 71: nit: empty line end of file

pkg/server/config.go
```diff
@@ -42,6 +61,36 @@ func (cfg *HandlerConfig) Validate() error {
 	return nil
 }
 
+// ValidateMappingConfig validates the handler config for mapping service after load.
+func (cfg *HandlerConfig) ValidateMappingConfig() error {
+	if err := cfg.Validate(); err != nil {
+		return err
```
sqin2019, line 67: Let's follow up in the next PR, we can join each error and return at the end, the original Validate func also need to be updated. here is an example - https://github.com/abcxyz/access-on-demand/blob/490e6a99855f0faf87a416cf221f1240bec003d5/apis/v1alpha1/validation.go#L79

pkg/server/config.go
```diff
@@ -42,6 +61,36 @@ func (cfg *HandlerConfig) Validate() error {
...
+	}
+
+	// For mapping server, we also require a failure topic ID.
+	if cfg.FailureTopicID == "" {
+		return fmt.Errorf("PMAP_FAILURE_TOPIC_ID is empty and require a value for mapping service")
+	}
+
+	if cfg.MappingConfig.DefaultResourceScope == "" {
+		return fmt.Errorf(`PMAP_RESOURCE_SCOPE is empty and require a value from one of the following format:\n
+			projects/{PROJECT_ID}\n
```
sqin2019, line 77: Just curious, is the "/n" regconized, how does it look in the error output?

sqin2019 (dismissed):
some nits.

pkg/server/config.go
```diff
@@ -16,17 +16,36 @@ package server
...
 // HandlerConfig defines the set over environment variables required
 // for running this application.
 type HandlerConfig struct {
 	Port           string `env:"PORT,default=8080"`
 	ProjectID      string `env:"PROJECT_ID,required"`
 	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
+	// FailureTopicID is optional for policy service
 	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
+	// MappingConfig is special config for mapping service
+	MappingConfig MappingConfig
```
yolocs, line 41: I think this should be the other way around. Define a `MappingHandlerConfig` that embeds `HandlerConfig`

pkg/server/config.go
```diff
@@ -16,17 +16,36 @@ package server
...
+}
+
+// MappingConfig defines the extra environment variables required
+// for running mapping service.
+type MappingConfig struct {
+	// DefaultResourceScope is required for mapping service
+	DefaultResourceScope string `env:"PMAP_RESOURCE_SCOPE"`
```
yolocs, line 48:
```suggestion
	DefaultResourceScope string `env:"PMAP_MAPPING_RESOURCE_SCOPE"`
```

pkg/server/config.go
```diff
@@ -42,6 +61,36 @@ func (cfg *HandlerConfig) Validate() error {
 	return nil
 }
 
+// ValidateMappingConfig validates the handler config for mapping service after load.
+func (cfg *HandlerConfig) ValidateMappingConfig() error {
```
yolocs, line 65: The `MappingHandlerConfig` should have its own `Validate` func and calls the `HandlerConfig.Validate`

pkg/server/config.go
yolocs, line 77: Your way of using literal string is problematic. Your output will include unnecessary tabs.

```suggestion
		allowedScopes := [
				"projects/[PROJECT_ID]",
				"projects/[PROJECT_NUMBER]",
				"folders/[FOLDER_NUMBER]",
				"organizations/[ORGNANIZATION_NUMBER]",
		]
		return fmt.Errorf("PMAP_MAPPING_RESOURCE_SCOPE is required in one of the formats: %v", allowedScopes)
```

yolocs: Define `allowedScopes` as a var on the top

pkg/server/config.go
```diff
@@ -42,6 +61,36 @@ func (cfg *HandlerConfig) Validate() error {
...
+	}
+
+	scope := strings.Split(cfg.MappingConfig.DefaultResourceScope, "/")[0]
+	if _, ok := SupportedResourceScope[scope]; !ok {
+		return fmt.Errorf(`PMAP_RESOURCE_SCOPE: %s doesn't have a valid value, the ResourceScope should be one of the following formats:\n
+		projects/{PROJECT_ID}\n
+		projects/{PROJECT_NUMBER}\n
+		folders/{FOLDER_NUMBER}\n
+		organizations/{ORGANIZATION_NUMBER}`, cfg.MappingConfig.DefaultResourceScope)
+	}
```
yolocs, line 90: No need to use a map
```suggestion
  switch scope {
  case "projects", "folders", "organizations":
    // continue
  default:
    return fmt.Errorf("PMAP_MAPPING_RESOURCE_SCOPE is required in one of the formats: %v", allowedScopes)
  }
```

yolocs (changes requested):

pkg/server/config.go
sailorlqh: Created #140

pkg/server/config.go
sqin2019: It will be used here too https://github.com/abcxyz/pmap/blob/main/pkg/mapping/processors/assetinventory.go#L250, might  be better to make it a var.

sailorlqh: I would also think using a map is better.

pkg/server/config.go
sailorlqh: If we agree to use map (see reason from the other conversation), I think we can keep allowedScopes in the validate function.

pkg/server/config.go
sailorlqh: Done.

pkg/server/config.go
```diff
@@ -16,19 +16,37 @@ package server
...
 // for running this application.
 type HandlerConfig struct {
 	Port           string `env:"PORT,default=8080"`
 	ProjectID      string `env:"PROJECT_ID,required"`
 	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
+	// FailureTopicID is optional for policy service
 	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
 }
 
+// MappingConfig defines the extra environment variables required
```
sqin2019, line 42: Nit: remove extra

pkg/server/config.go
```diff
@@ -16,19 +16,37 @@ package server
...
+// for running mapping service.
+type MappingHandlerConfig struct {
+	// DefaultResourceScope is required for mapping service
+	DefaultResourceScope string `env:"PMAP_MAPPING_RESOURCE_SCOPE"`
```
sqin2019, line 46: Now we can add `env:"PMAP_MAPPING_RESOURCE_SCOPE,required"`

terraform/e2e/variables.tf
```diff
@@ -59,7 +59,7 @@ variable "bigquery_table_delete_protection" {
 
 variable "pmap_mapping_resource_scope" {
   description = <<EOT
-        "This is the scope for where the resources resides in.
+        "The scope for where the resources resides in.
```
sqin2019, line 62: I meant we dont need the double quote here.

terraform/modules/pmap-service/variables.tf
```diff
@@ -76,7 +76,7 @@ variable "gcs_events_filter" {
 
 variable "pmap_mapping_resource_scope" {
   description = <<EOT
-        "This is the scope for where the resources resides in.
+        "The scope for where the resources resides in.
```
sqin2019, line 79: And here.

sqin2019 (approved):

pkg/server/config.go
```diff
@@ -16,19 +16,37 @@ package server
...
+	// FailureTopicID is optional for policy service
 	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
 }
 
+// MappingConfig defines the environment variables required
+// for running mapping service.
+type MappingHandlerConfig struct {
+	// DefaultResourceScope is required for mapping service
+	DefaultResourceScope string `env:"PMAP_MAPPING_RESOURCE_SCOPE,required"`
+	HandlerConfig        HandlerConfig
```
yolocs, line 47: You can embed HandlerConfig instead. Accessing fields would be easier.

```go
type MappingHandlerConfig struct {
  HandlerConfig
  DefaultResourceScope string `env:"PMAP_MAPPING_RESOURCE_SCOPE,required"`
}
```

pkg/server/config.go
yolocs: No, the processor package (where asset inventory processor lives) should not take a dependency on the server package (where the config lives). They should be decoupled.

pkg/server/config.go
yolocs: See my other comment. And again, your usage of ` is wrong. See the difference here https://go.dev/play/p/Ag73GyblH4l

pkg/server/config.go
yolocs: Also should the env var be `PMAP_MAPPING_DEFAULT_RESOURCE_SCOPE`?

pkg/server/config.go
```diff
@@ -16,19 +16,37 @@ package server
...
 	ProjectID      string `env:"PROJECT_ID,required"`
 	SuccessTopicID string `env:"PMAP_SUCCESS_TOPIC_ID,required"`
+	// FailureTopicID is optional for policy service
 	FailureTopicID string `env:"PMAP_FAILURE_TOPIC_ID"`
 }
 
+// MappingConfig defines the environment variables required
+// for running mapping service.
+type MappingHandlerConfig struct {
+	// DefaultResourceScope is required for mapping service
```
yolocs, line 45:
```suggestion
	// DefaultResourceScope is the default resource scope to search resources.
	// This is only used for global resources such as GCS bucket.
```

pkg/server/config.go
```diff
@@ -80,3 +128,22 @@ func (cfg *HandlerConfig) ToFlags(set *cli.FlagSet) *cli.FlagSet {
...
+	f.StringVar(&cli.StringVar{
+		Name:    "default-resource-scope",
+		Target:  &cfg.DefaultResourceScope,
+		EnvVar:  "PMAP_MAPPING_RESOURCE_SCOPE",
+		Example: "{projects/test-project-id}",
+		Usage: `The default scope to search for resources. Format: 
+		projects/{PROJECT_ID}
+		projects/{PROJECT_NUMBER}
+		folders/{FOLDER_NUMBER}
+		organizations/{ORGANIZATION_NUMBER}`,
```
yolocs, line 146: Define `allowedScopes` in the file scope so you can use it here as well. Again, you usage of ` is wrong.

terraform/e2e/variables.tf
```diff
@@ -56,3 +56,16 @@ variable "bigquery_table_delete_protection" {
...
+  description = <<EOT
+        "The scope for where the resources resides in.
+        Options can be one of the following:
+        projects/{PROJECT_ID}
+        projects/{PROJECT_NUMBER}
+        folders/{FOLDER_NUMBER}
+        organizations/{ORGANIZATION_NUMBER}"
+    EOT
+  type        = string
+  default     = ""
```
yolocs, line 70: If this is required, why do we still default to empty?

pkg/cli/mapping_server.go
```diff
@@ -93,24 +93,19 @@ func (c *MappingServerCommand) RunUnstarted(ctx context.Context, args []string)
...
-
-	pubsubClient, err := pubsub.NewClient(ctx, c.cfg.ProjectID)
+	pubsubClient, err := pubsub.NewClient(ctx, c.cfg.HandlerConfig.ProjectID)
 	if err != nil {
 		return nil, nil, closer, fmt.Errorf("failed to create pubsub client: %w", err)
 	}
 	closer = multicloser.Append(closer, pubsubClient.Close)
 
-	successTopic := pubsubClient.Topic(c.cfg.SuccessTopicID)
+	successTopic := pubsubClient.Topic(c.cfg.HandlerConfig.SuccessTopicID)
```
yolocs, line 106: If you embed HandlerConfig, then these lines don't have to change.

pkg/mapping/processors/assetinventory.go
```diff
@@ -241,22 +242,24 @@ func mergeAnnotations(annos1, annos2 *structpb.Struct) (*structpb.Struct, error)
...
+// parseScope gets "project/folder/orgnization" from "ResourceName" follows format here:https://cloud.google.com/asset-inventory/docs/resource-name-format.
+// Return empty string for resources such as GCS bucket won't include "project/folder/orgnization" info in its "ResourceName".
+func parseScope(resourceName string) (string, error) {
 	s := strings.Split(resourceName, "/")
-	project := ""
+	scopePrefix := ""
+	scope := ""
 	for i, e := range s {
-		if e == "projects" {
+		if _, ok := server.SupportedResourceScope[e]; ok {
```
yolocs, line 252: This is bad. Don't depend on `server` package. Repeating projects/folders/organizations is ok

yolocs (changes requested):

terraform/e2e/variables.tf
sailorlqh: Because mapping and policy uses the same module. https://github.com/abcxyz/pmap/blob/main/terraform/e2e/main.tf#L30-L59

If there is no default value here, the policy module would fail. Now I default the value to `projects/project_id`

yolocs: I see. Thanks!

yolocs (approved):

Issue #141, opened 2023-07-06
Add integ test case on getting GCS bucket info from other project (pmap-dev)


sailorlqh:
As #137 merged, we now can customize scopes. Currently the e2e test only test GCS buckets within the same project. 
We can to add a test case to test on GCS bucket that is from a different project (maybe from pmap-dev).

PR #142, opened 2023-07-10
feat: add e2e test converage to test pmap customized scope

sailorlqh:
fix: #141

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,8 @@ env:
   INTEG_TEST_POLICY_UPSTREAM_TOPIC: 'policy-gcs'
   INTEG_TEST_POLICY_DOWNSTREAM_TOPIC: 'policy-bigquery'
   INTEG_TEST_STATIC_GCS_BUCKET: 'pmap-static-ci-bucket-7faa'
+  INTEG_TEST_STATIC_AR_REPO: 'pmap-static-ci-artifact-registry-repo-7faa'
+  INTEG_TEST_MAPPING_DEFAULT_RESOURCE_SCOPE: 'folders/1082218530939'
 
 concurrency:
   group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -125,14 +125,14 @@ func TestMappingEventHandling(t *testing.T) {
 		wantProcessErrSubStr string
 	}{
 		{
-			name:              "mapping_success_event_unscoped_resource",
-			resourceName:      fmt.Sprintf("//storage.googleapis.com/%s", cfg.StaticGCSBucket),
+			name:              "mapping_success_event_scoped_resource",
+			resourceName:      fmt.Sprintf("//artifactregistry.googleapis.com/projects/%s/locations/us-central1/repositories/%s", cfg.ProjectID, cfg.StaticARRepo),
 			bigqueryTable:     cfg.MappingTableID,
 			wantCAISProcessed: true,
 			wantResourceMapping: &v1alpha1.ResourceMapping{
 				Resource: &v1alpha1.Resource{
 					Provider: "gcp",
-					Name:     fmt.Sprintf("//storage.googleapis.com/%s", cfg.StaticGCSBucket),
+					Name:     fmt.Sprintf("//artifactregistry.googleapis.com/projects/%s/locations/us-central1/repositories/%s", cfg.ProjectID, cfg.StaticARRepo),
 				},
 				Contacts: &v1alpha1.Contacts{Email: []string{"group@example.com"}},
 			},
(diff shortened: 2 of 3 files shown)


test/integration/main_test.go
```diff
@@ -125,14 +125,36 @@ func TestMappingEventHandling(t *testing.T) {
...
+				Commit:                     testGithubCommitValue,
+				Workflow:                   testWorkflowValue,
+				WorkflowSha:                testWorkflowShaValue,
+				WorkflowTriggeredTimestamp: testParseTime(t, testWorkflowTriggeredTimeValue),
+				WorkflowRunId:              testWorkflowRunID,
+				WorkflowRunAttempt:         1,
+			},
+		},
+		{
+			name:              "mapping_success_event_dev",
```
sailorlqh, line 150: I considered about combining these two test into one that only obtain Bucket info from pmap-dev project as pmap-dev and pmap-ci are under the same folder, and the default scope now is `folders`. But I thought having two tests here can help us make sure we can not only get resource info from the same project, but also from other project under the same folder.
Please let me know if you think this is redundant.

sqin2019: I think it is unnecessary when your scope is folder, cause getting the resource info in a scope is Asset Inventory's job and asset inventory should make sure of that. Do we have a non-GCS resource test?

.github/workflows/ci.yml
```diff
@@ -33,7 +33,9 @@ env:
   INTEG_TEST_MAPPING_DOWNSTREAM_FAILURE_TOPIC: 'mapping-failure-bigquery'
   INTEG_TEST_POLICY_UPSTREAM_TOPIC: 'policy-gcs'
   INTEG_TEST_POLICY_DOWNSTREAM_TOPIC: 'policy-bigquery'
-  INTEG_TEST_STATIC_GCS_BUCKET: 'pmap-static-ci-bucket-7faa'
+  INTEG_TEST_STATIC_GCS_BUCKET_CI: 'pmap-static-ci-bucket-7faa'
+  INTEG_TEST_STATIC_GCS_BUCKET_DEV: 'pmap-dev'
```
capri-xiyue, line 37: Maybe create a dedicated gcs bucket for this e2e test and use specific naming.

capri-xiyue (approved):

test/integration/main_test.go
```diff
@@ -125,14 +125,38 @@ func TestMappingEventHandling(t *testing.T) {
...
+				RepoName:                   testGithubRepoValue,
+				Commit:                     testGithubCommitValue,
+				Workflow:                   testWorkflowValue,
+				WorkflowSha:                testWorkflowShaValue,
+				WorkflowTriggeredTimestamp: testParseTime(t, testWorkflowTriggeredTimeValue),
+				WorkflowRunId:              testWorkflowRunID,
+				WorkflowRunAttempt:         1,
+			},
+		},
+		// This test focus on testing resources don't have projects/project_id in it's name such as GCS bucket.
```
capri-xiyue, line 150: Make it clear that it also tests the customized scope

test/integration/main_test.go
```diff
@@ -125,14 +125,38 @@ func TestMappingEventHandling(t *testing.T) {
 		wantProcessErrSubStr string
 	}{
 		{
-			name:              "mapping_success_event",
-			resourceName:      fmt.Sprintf("//storage.googleapis.com/%s", cfg.GCSStaticBucket),
+			// This test focus on testing resources that has projects/project_id in it's name.
+			name:              "mapping_success_event_ci",
```
capri-xiyue, line 129: I think you'd better name the test case as `mapping_success_event` and `mapping_success_event_with_customized_scope`

.github/workflows/ci.yml
sqin2019: Also, we can create it in ci project.

.github/workflows/ci.yml
```diff
@@ -33,7 +33,9 @@ env:
   INTEG_TEST_MAPPING_DOWNSTREAM_FAILURE_TOPIC: 'mapping-failure-bigquery'
   INTEG_TEST_POLICY_UPSTREAM_TOPIC: 'policy-gcs'
   INTEG_TEST_POLICY_DOWNSTREAM_TOPIC: 'policy-bigquery'
-  INTEG_TEST_STATIC_GCS_BUCKET: 'pmap-static-ci-bucket-7faa'
+  INTEG_TEST_STATIC_ARTIFACT_REPO_CI: 'ci-images'
```
capri-xiyue, line 36: I'd prefer you use static resources specifically for e2e test.

.github/workflows/ci.yml
```diff
@@ -33,7 +33,9 @@ env:
...
+  INTEG_TEST_STATIC_GCS_BUCKET_DEV: 'pmap-dev'
```
capri-xiyue, line 37: The `dev` naming is confusing here.

sailorlqh:
/Hold until [PR#146](https://github.com/abcxyz/pmap/pull/146) for creating static resources is merged and applied.

test/integration/main_test.go
sqin2019: We can add another test that asset inventory search fails, for example a GCS with wrong resource scope. This could provide good test coverage for custom scope, and we don't need to create resource in dev.

capri-xiyue (approved):

capri-xiyue (approved):

yolocs (approved):

PR #146, opened 2023-07-10
fix: add static resources for e2e test

sailorlqh:


terraform/e2e/main.tf
```diff
@@ -59,3 +59,34 @@ module "policy_service" {
   oidc_service_account = module.common_infra.oidc_service_account
   gcs_events_filter    = var.policy_gcs_events_filter
 }
+
+resource "random_id" "default" {
```
capri-xiyue, line 63: Why do you need this for e2e env?

terraform/ci/main.tf
```diff
@@ -30,13 +30,14 @@ resource "random_id" "default" {
 
 
 locals {
-  static_bucket_name = "pmap-static-ci-bucket-${random_id.default.hex}"
+  ci_static_bucket_name        = "pmap-static-ci-bucket-${random_id.default.hex}"
```
capri-xiyue, line 33: No need to add `ci` as prefix as this is terraform/ci/main.tf

terraform/e2e/main.tf
sailorlqh: I thought a random prefix can't do any harm. So you prefer to just keep the name as `pmap-static-dev-bucket`?

terraform/ci/main.tf
sailorlqh: Removed.

terraform/e2e/main.tf
capri-xiyue: I mean why you need to add all these resources like storage bucket in e2e env

capri-xiyue: I got it. You want to test the customized scope test case  in the `pmap-dev`, right?Please add a comment to explain what all these resources are for and later clean the static bucket in `pmap-ci` with another PR.

terraform/e2e/main.tf
```diff
@@ -59,3 +59,34 @@ module "policy_service" {
...
+  byte_length = 2
+}
+
+locals {
+  dev_static_bucket_name = "pmap-static-dev-bucket-${random_id.default.hex}"
```
capri-xiyue, line 68: use `static_bucket_name        = "pmap-static-ci-bucket-${random_id.default.hex}"`

terraform/e2e/main.tf
sailorlqh: Because we want the e2e test to get resource from another project to make sure the customize resource scope functionality is working. So we create a static bucket in pamp-dev project.

terraform/e2e/main.tf
sailorlqh: Done.

capri-xiyue: You are still using the previous variable name.

sailorlqh: Done.

terraform/e2e/main.tf
sqin2019: I don't think e2e module is for this purpose, it is for us and customers to set up infrastructure required for pmap, it does not make sense to create ci resource in this module. 

I wonder why we cannot test custom scope in pmap-ci?

sailorlqh: Because for most resources, they have `projects/{project_id}` within its name, and we can parse that from the name and use it as the scope. However, GCS bucket's name is special, it doesn't have `projects/{project_id}` in its name. So when getting info from GCS buckets, the search scope is defined by `defaultScope`. And if user didn't define a `defaultScope`, `defaultScope` will be set to `project/{project_id}`, To make sure `defaultScope` is working when set to `folders` or `organization`, we need to test on some resources that's not from the same project.

sqin2019: Please correct me, I thought we don't set the default anymore and users need to provide one?

sqin2019: This should help? https://github.com/abcxyz/pmap/pull/142/files#r1259018960

capri-xiyue: +1 We don't set the default anymore and users need to provide one, right? `And if user didn't define a defaultScope, defaultScope will be set to project/{project_id}`, this logic is not correct.

sailorlqh: From the discussion, I am thinking only create a non-GCS resource in ci project, and use existing pmap-dev GCS bucket to test `defaultResouceScope`.

yolocs: Agree with @sqin2019 that the e2e module is not for this purpose. The test bucket should only be created in the ci module.

yolocs (changes requested):

terraform/ci/main.tf
```diff
@@ -55,3 +56,29 @@ resource "google_storage_bucket_iam_member" "static_bucket_object_viewer" {
   role   = "roles/storage.objectViewer"
   member = "serviceAccount:${module.common_infra.run_service_account}"
 }
+
+# Add a static artifact repo for e2e test
+resource "google_artifact_registry_repository" "integ_test_deicated_artifact_repo" {
```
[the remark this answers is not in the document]
sailorlqh: The reason for using AR repo is here: https://github.com/abcxyz/pmap/pull/142#discussion_r1258897814.
We want to test a GCS resource outside of the current project, and a non-GCS resource within the same project.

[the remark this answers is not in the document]
yolocs: I'm not following. 

1. GCS resource is unscoped so even if it's in the same project, we already have the right test coverage.
2. Non-GCS resource in the same project - don't we already test that with a BQ dataset or something?

sailorlqh:
Close this PR as creating resource in e2e module is no longer needed.

PR #148, opened 2023-07-11
fix: create static scoped non-GCS bucket resource (AR repo) in CI for integration test

sailorlqh:
Create a artifact registry repo as a static scoped non-GCS bucket resource in CI for integration test.

### DIFF
--- a/terraform/ci/main.tf
+++ b/terraform/ci/main.tf
@@ -30,9 +30,12 @@ resource "random_id" "default" {
 
 
 locals {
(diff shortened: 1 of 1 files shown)


yolocs (commented):
Why we need this again?

capri-xiyue:
> Why we need this again?

We need one non-gcs bucket resource to test the scope parsing logic work

yolocs (approved):

Issue #149, opened 2023-07-11
proto unmarshal error will keep retrying in the pmap service


capri-xiyue:
### TL;DR

Currently, with this https://github.com/abcxyz/pmap/blob/9619165747d9f4c3bca38973a836ac87c2a0094d/pkg/server/event_handler.go#L333, if we fail to unmarshal the gcs object, it should be a user facing error and it should go to the failure table. Currently the behavior is it keeps retrying in the pmap service.

### Environment Details

```markdown
all
```

PR #150, opened 2023-07-12
Fix: update errors from protoutil.FromYAML() as user facing error

sailorlqh:
fix: #149


pkg/server/event_handler_test.go
```diff
@@ -613,11 +620,20 @@ func (m *testMessenger) Send(_ context.Context, data []byte, attr map[string]str
...
 	}
-	err := protojson.Unmarshal(data, m.gotPmapEvent)
-	if err != nil {
-		return fmt.Errorf("failed to unmarshal to PmapEvent: %w", err)
+
+	// len(data) can be zero when generatePmapEventBytes() fails.
+	// if len(data) is zero, there is no point to Unmarshal data,
+	// the unmarshal here is to help test generatePmapEventBytes() generated
+	// the right eventBytes.
+	if len(data) != 0 {
```
capri-xiyue, line 628: Why do you need this?

pkg/server/event_handler.go
```diff
@@ -331,7 +332,7 @@ func (h *EventHandler[T, P]) getGCSObjectProto(ctx context.Context, objAttrs map
 	// Convert the object yaml bytes into a proto message wrapper.
 	p := P(new(T))
```
capri-xiyue, line 333: I'm wondering maybe we should refactor the `getGCSObjectProto` method to just get gcs object bytes and we have another function to convert the object bytes to proto. Currently it is a little bit weird that the get gcs object will include pmaperrors. It's better that the function has just one single responsibility.

pkg/server/event_handler_test.go
sailorlqh: Because when there is some error with yaml parsing, the data will be nil (have length of 0), and this will cause `protojson.Unmarshal()` to throw error.
See: https://go.dev/play/p/hHa3y_AC36c

capri-xiyue: Why the data will be nil, I thought we will write some failure info into failure table.

sailorlqh: Discussed offline and explained why, updated the comment to better explain the usage of this function.

pkg/server/event_handler.go
sailorlqh: Decoupled.

pkg/server/event_handler.go
```diff
@@ -266,11 +266,21 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 
 func (h *EventHandler[T, P]) generatePmapEventBytes(ctx context.Context, m pubsub.Message) ([]byte, error) {
 	// Get the GCS object as a proto message given GCS notification information.
-	p, err := h.getGCSObjectProto(ctx, m.Attributes)
+	var processErr error
+
+	b, err := h.getGCSObjectBytes(ctx, m.Attributes)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get GCS object: %w", err)
+		return nil, errors.Join(processErr, fmt.Errorf("failed to get GCS object: %w", err))
```
yolocs, line 273: `processErr` is nil at this point. What's the point of joining it?

pkg/server/event_handler.go
```diff
@@ -266,11 +266,21 @@ func (h *EventHandler[T, P]) Handle(ctx context.Context, m pubsub.Message) error
 
 func (h *EventHandler[T, P]) generatePmapEventBytes(ctx context.Context, m pubsub.Message) ([]byte, error) {
 	// Get the GCS object as a proto message given GCS notification information.
-	p, err := h.getGCSObjectProto(ctx, m.Attributes)
```
capri-xiyue, line 269: Do we need include get gcs object error as part of the processErr, I thought the process error is just for processors?

pkg/server/event_handler_test.go
yolocs: Would pubsub fail if data is nil?

I think the comment here could be as simple as "Don't record gotPmapEvent when data is nil". The current comment requires too much context from other places. I can't even find where it defines `generatePmapEventBytes`

pkg/server/event_handler.go
yolocs: One the contrary, we should join errors on line 293.

pkg/server/event_handler_test.go
capri-xiyue: Confirmed with @sailorlqh , pubsub won't fail. Here we need to do unmarshal in UT is because UT needs to verify the send pamp event.

sailorlqh: pubsub won't fail. See [here](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage#:~:text=A%20message%20that%20is%20published,differently%20depending%20on%20the%20language.)

If data is empty, as long as we have one attribute, publish won't fail. And we do have attribute here.

pkg/server/event_handler.go
sailorlqh: Done.

pkg/server/event_handler.go
sailorlqh: My bad, removed.

yolocs: Can you move it to somewhere near line 283?

sailorlqh: Moved to line 282.

pkg/server/event_handler_test.go
```diff
@@ -613,11 +620,19 @@ func (m *testMessenger) Send(_ context.Context, data []byte, attr map[string]str
...
-	err := protojson.Unmarshal(data, m.gotPmapEvent)
-	if err != nil {
-		return fmt.Errorf("failed to unmarshal to PmapEvent: %w", err)
+
+	// Don't record gotPmapEvent when data is nil.
+	// Also protojson.Unmarshal will return err
+	// when unmarshal an empty byte slice
+	if len(data) != 0 {
+		err := protojson.Unmarshal(data, m.gotPmapEvent)
+		if err != nil {
```
yolocs, line 629:
```suggestion
		if err := protojson.Unmarshal(data, m.gotPmapEvent); err != nil {
```

yolocs (approved):

capri-xiyue (dismissed):

capri-xiyue (approved):

yolocs (approved):

PR #152, opened 2023-07-18
feat: added dlq pubsub for gcs push subscription

capri-xiyue:
part of #118

### DIFF
--- a/terraform/modules/pmap-service/variables.tf
+++ b/terraform/modules/pmap-service/variables.tf
@@ -86,3 +86,9 @@ variable "pmap_specific_envvars" {
     EOT
   default     = {}
 }
+
+variable "enable_dead_lettering" {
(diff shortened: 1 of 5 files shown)


terraform/modules/pmap-service/variables.tf
```diff
@@ -86,3 +86,9 @@ variable "pmap_specific_envvars" {
     EOT
   default     = {}
 }
+
+variable "enable_dead_lettering" {
+  description = "Enables the creation of a dead letter topic for the pub/sub push subscription."
+  type        = bool
+  default     = true
```
sqin2019, line 93: Do we want the default to be true? for example for ci, we probably don't need it since the dlq will be deleted every ci run. either change the default or we can update the ci to set it to false?

sqin2019 (approved):

PR #158, opened 2023-07-26
feat: add prober code

sailorlqh:
add go code for prober, as part of the fix of #157.

Probing for policy service, other resource setup like cloud run job, and goreleaser update will in following PRs.

### DIFF
--- a/prober/config.go
+++ b/prober/config.go
@@ -23,14 +23,13 @@ import (
 )
 
 type config struct {
-	ProjectID                string        `env:"PROBER_PROJECT_ID,required"`
-	GCSBucketID              string        `env:"PROBER_BUCKET_ID,required"`
-	BigQueryDataSetID        string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
-	MappingTableID           string        `env:"PROBER_MAPPING_TABLE_ID,required"`
-	QueryRetryWaitDuration   time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
-	QueryRetryLimit          uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
-	ProberMappingServiceName string        `env:"PROBER_MAPPING_SERVICE_NAME,required"`
-	ProberMappingFilePrefix  string        `env:"PROBER_MAPPING_FILE_PREFIX,required"`
+	ProjectID                    string        `env:"PROBER_PROJECT_ID,required"`
+	GCSBucketID                  string        `env:"PROBER_BUCKET_ID,required"`
+	BigQueryDataSetID            string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
+	MappingTableID               string        `env:"PROBER_MAPPING_TABLE_ID,required"`
+	QueryRetryWaitDuration       time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
+	QueryRetryLimit              uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
+	ProberMappingGCSBucketPrefix string        `env:"PROBER_MAPPING_BUCKET_PREFIX,required"`
 }
 
 func newTestConfig(ctx context.Context) (*config, error) {
--- a/prober/config.go
+++ b/prober/config.go
@@ -1,0 +1,40 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/sethvargo/go-envconfig"
+)
+
+type config struct {
+	ProjectID              string        `env:"PROBER_PROJECT_ID,required"`
+	GCSBucketID            string        `env:"PROBER_BUCKET_ID,required"`
+	BigQueryDataSetID      string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
+	MappingTableID         string        `env:"PROBER_MAPPING_TABLE_ID,required"`
+	QueryRetryWaitDuration time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
+	QueryRetryLimit        uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
+}
+
+func newTestConfig(ctx context.Context) (*config, error) {
+	var c config
+	if err := envconfig.ProcessWith(ctx, &c, envconfig.OsLookuper()); err != nil {
+		return nil, fmt.Errorf("failed to process environment: %w", err)
+	}
+	return &c, nil
+}
--- a/prober/config.go
+++ b/prober/config.go
@@ -23,12 +23,14 @@ import (
 )
 
 type config struct {
-	ProjectID              string        `env:"PROBER_PROJECT_ID,required"`
-	GCSBucketID            string        `env:"PROBER_BUCKET_ID,required"`
-	BigQueryDataSetID      string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
-	MappingTableID         string        `env:"PROBER_MAPPING_TABLE_ID,required"`
-	QueryRetryWaitDuration time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
-	QueryRetryLimit        uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
+	ProjectID                string        `env:"PROBER_PROJECT_ID,required"`
+	GCSBucketID              string        `env:"PROBER_BUCKET_ID,required"`
+	BigQueryDataSetID        string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
+	MappingTableID           string        `env:"PROBER_MAPPING_TABLE_ID,required"`
+	QueryRetryWaitDuration   time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
+	QueryRetryLimit          uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
+	ProberMappingServiceName string        `env:"PROBER_MAPPING_SERVICE_NAME,required"`
+	ProberMappingFilePrefix  string        `env:"PROBER_MAPPING_FILE_PREFIX,required"`
 }
 
 func newTestConfig(ctx context.Context) (*config, error) {
--- a/prober/config.go
+++ b/prober/config.go
@@ -29,7 +29,7 @@ type config struct {
 	MappingTableID               string        `env:"PROBER_MAPPING_TABLE_ID,required"`
 	QueryRetryWaitDuration       time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
 	QueryRetryLimit              uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
-	ProberMappingGCSBucketPrefix string        `env:"PROBER_MAPPING_BUCKET_PREFIX,required"`
+	ProberMappingGCSBucketPrefix string        `env:"PROBER_MAPPING_GCS_BUCKET_PREFIX,required"`
 }
 
 func newTestConfig(ctx context.Context) (*config, error) {
--- a/prober/main.go
+++ b/prober/main.go
@@ -45,6 +45,7 @@ const (
 	proberWorkflowRunAttempt   = "1"
 	proberGCSNamePrefix        = "//storage.googleapis.com"
 	proberMappingTraceIDPrefix = "prober-mapping"
+	proberMappingFilePrefix    = "prober-file"
 	proberResourceProvider     = "gcp"
 	proberResourceContact      = "pmap-prober@abcxyz.dev"
 )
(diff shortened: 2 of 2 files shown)


prober/config.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+
+	"github.com/sethvargo/go-envconfig"
+)
+
+type config struct {
+	ProjectID              string        `env:"PROBER_PROJECT_ID,required"`
+	GCSBucketID            string        `env:"PROBER_BUCKET_ID,required"`
+	BigQueryDataSetID      string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
+	MappingTableID         string        `env:"PROBER_MAPPING_TABLE_ID,required"`
```
capri-xiyue, line 29: I'd suggest implementing prober for mapping and policy in one PR so that we can figure out what the structure looks like

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+	if err := wc.Close(); err != nil {
+		return fmt.Errorf("failed to close writer: %w", err)
+	}
+
+	return nil
+}
+
+// getFirstMatchedBQEntryWithRetries query BigQuery table to find and return the matching entry.
+// If no result is found, query will be retried with the retry config.
+func getFirstMatchedBQEntryWithRetries(ctx context.Context, bqQuery *bigquery.Query, cfg *config) (string, error) {
```
capri-xiyue, line 210: I think this is just duplication of e2e tests. We need to consider to refactor it to get rid of duplication

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+		log.Print("Matching entry not found, retrying...")
+		return retry.RetryableError(fmt.Errorf("no matching pmap event found in bigquery after timeout"))
+	}); err != nil {
+		return "", fmt.Errorf("retry failed: %w", err)
+	}
+	return entry, nil
+}
+
+// queryBQEntries queries the BQ and checks if a bigqury entry matched the query exists or not and return the results.
+func queryBQEntries(ctx context.Context, query *bigquery.Query) ([]string, error) {
```
capri-xiyue, line 233: Same here, need refactor

prober/config.go
sailorlqh: I thought it's better to divide PR into smallest possible pieces so I didn't include the policy part. 
For probing policy service, only a new field `PolicyTableID` needs to be added.

prober/main.go
sailorlqh: I don't think making prober depends on package `integration` is a good idea. To do that, we need to export the functions, and plus, prober doesn't need stuffs like `testing.TB` but that's required by the integration test.

prober/config.go
capri-xiyue: I doubt whether we should put mapping and policy prober into one main function/config.

prober/main.go
capri-xiyue: I don't mean you make prober depends on package integration and just simple export `getFirstMatchedBQEntryWithRetries`. Further refactoring could be needed to avoid the duplication. My point is to get rid of duplication. There are other ways to achieve it instead of just exporting the `getFirstMatchedBQEntryWithRetries`. For example, you can extract GCS and BQ utility function

sailorlqh:
/Hold until PR#159 for refactoring the common code is merged.

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// create a global bigquery client
+	bc, err := bigquery.NewClient(ctx, c.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to create bigquery client: %w", err)
+	}
+	bqClient = bc
+	defer bc.Close()
+
+	ts := time.Now().Format(time.RFC3339)
```
yolocs, line 91: User `Unix()`?

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+	log.Printf("using traceID: %s", traceID)
+	data := []byte(fmt.Sprintf(`
+resource:
+  name: //storage.googleapis.com/%s
+  provider: gcp
+annotations:
+  traceID: %s
+contacts:
+  email:
+  - prober@pmap.com
```
yolocs, line 118: nit: define the pattern as a const

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+`, cfg.GCSBucketID, traceID))
+
+	filepath := fmt.Sprintf("mapping/prober-files/timestamp-%s", timestamp)
```
yolocs, line 121: nit: define the pattern as a const

prober/main.go
yolocs, line 118: nit: maybe `pmap-prober@abcxyz.dev`

yolocs (changes requested):

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// TODO: add probePolicy to probe policy service
+	var probeErr error
+	if err := probeMapping(ctx, ts); err != nil {
+		probeErr = errors.Join(probeErr, fmt.Errorf("prober failed for mapping service: %w", err))
+	} else {
+		log.Print("Mapping probe succeed")
+	}
+
+	return probeErr
```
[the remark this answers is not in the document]
sailorlqh: I was intended to use one prober to probe both mapping and policy service, and return the errors together, so I used `errors.Join()` here.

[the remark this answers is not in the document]
yolocs: I see. A note will be helpful

prober/main.go
```diff
@@ -0,0 +1,266 @@
...
+	bc, err := bigquery.NewClient(ctx, c.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to create bigquery client: %w", err)
+	}
+	bqClient = bc
+	defer bc.Close()
+
+	ts := time.Now().Format(time.RFC3339)
+
+	// TODO: add probePolicy to probe policy service
```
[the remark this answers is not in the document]
sailorlqh: It's probePolicy() to probe policy service.

prober/main.go
sailorlqh: I think either `Unix()` or `time.RFC3339` fits the requirement , but `time.RFC3339` can be more human readable.

prober/config.go
sailorlqh: If we want one prober to probe both mapping and policy service, this is necessary.

prober/main.go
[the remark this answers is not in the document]
sailorlqh: Note added.

prober/main.go
```diff
@@ -0,0 +1,210 @@
...
+const (
+	proberGithubCommitValue    = "prober-test-github-commit"
+	proberGithubRepoValue      = "prober-test-github-repo"
+	proberWorkflowValue        = "prober-test-workflow"
+	proberWorkflowShaValue     = "prober-test-workflow-sha"
+	proberWorkflowRunID        = "prober-test-workflow-id"
+	proberWorkflowRunAttempt   = "1"
+	proberGCSNamePrefix        = "//storage.googleapis.com"
+	proberMappingTraceIDPrefix = "prober-mapping"
+	proberMappingFilePrefix    = "mapping/prober-files/timestamp"
```
yolocs, line 47: This should probably be configurable. It needs to match the values here https://github.com/abcxyz/pmap/blob/main/terraform/e2e/main.tf#L15-L18

prober/main.go
```diff
@@ -0,0 +1,210 @@
...
+	ts := time.Now().Format(time.RFC3339)
+
+	// TODO: add probePolicy() to probe policy service
+	// errors.Join() is used here so we can join the errors from probePolicy and return them together.
+	var probeErr error
+	if err := probeMapping(ctx, ts); err != nil {
+		probeErr = errors.Join(probeErr, fmt.Errorf("prober failed for mapping service: %w", err))
+	} else {
+		logging.Info("Mapping probe succeed.")
+	}
```
yolocs, line 104:
```suggestion
	var probeErr error
	if err := probeMapping(ctx, ts); err != nil {
		probeErr = errors.Join(probeErr, fmt.Errorf("prober failed for mapping service: %w", err))
	}
	// TODO: add probePolicy() to probe policy service.
	// if err := probePolicy(ctx, ts); err != nil {
	//   probeErr = errors.Join(probeErr, fmt.Errorf("prober failed for policy service: %w", err))
	// }
	if probeErr != nil {
		logging.Info("Mapping probe succeed.")
	}
```

yolocs (commented):
Have you manually tried probing to see if it works?

prober/main.go
capri-xiyue: +1

prober/main.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// create a global bigquery client
+	bc, err := bigquery.NewClient(ctx, c.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to create bigquery client: %w", err)
+	}
+	bqClient = bc
+	defer bc.Close()
+
+	ts := time.Now().Format(time.RFC3339)
```
capri-xiyue, line 95: Can this ts be created in `probeMapping` function. I don't see the point now the timestamp should be a parameter

prober/main.go
```diff
@@ -0,0 +1,210 @@
...
+		Workflow:           proberWorkflowValue,
+		WorkflowSha:        proberWorkflowShaValue,
+		WorkflowRunId:      proberWorkflowRunID,
+		WorkflowRunAttempt: 1,
+	}
+
+	if diff := cmp.Diff(wantGithubSource, gotPmapEvent.GetGithubSource(), cmpOpts...); diff != "" {
+		diffErr = errors.Join(diffErr, fmt.Errorf("githubSource unexpected diff (-want, +got):\n%s", diff))
+	}
+
```
capri-xiyue, line 186: I think we need to check some CAIS annotation exist to see whether CAIS asset inventory works or not.

prober/main.go
```diff
@@ -0,0 +1,210 @@
...
+	wantGithubSource := &v1alpha1.GitHubSource{
+		RepoName:           proberGithubRepoValue,
+		Commit:             proberGithubCommitValue,
+		Workflow:           proberWorkflowValue,
+		WorkflowSha:        proberWorkflowShaValue,
+		WorkflowRunId:      proberWorkflowRunID,
+		WorkflowRunAttempt: 1,
+	}
+
+	if diff := cmp.Diff(wantGithubSource, gotPmapEvent.GetGithubSource(), cmpOpts...); diff != "" {
```
capri-xiyue, line 183: Why this use the same `cmpOpts` as that is used in resourceMapping?

capri-xiyue (changes requested):

prober/main.go
sailorlqh: I wanted the mapping and policy service to have the same timestamp, so it's more consistant.

prober/main.go
sailorlqh: Updated.

prober/main.go
sailorlqh: Added.

prober/main.go
sailorlqh: Updated.

prober/main.go
```diff
@@ -0,0 +1,225 @@
...
+
+	if diff := cmp.Diff(wantGithubSource, gotPmapEvent.GetGithubSource(), protocmp.Transform()); diff != "" {
+		diffErr = errors.Join(diffErr, fmt.Errorf("githubSource unexpected diff (-want, +got):\n%s", diff))
+	}
+
+	// check CAIS annotation exist to make sure CAIS is working
+	if _, ok := resourceMapping.GetAnnotations().GetFields()[v1alpha1.AnnotationKeyAssetInfo].GetStructValue().AsMap()["ancestors"]; !ok {
+		diffErr = errors.Join(diffErr, fmt.Errorf("ancestors is blank in resourcemapping.annotations"))
+	}
+	if _, ok := resourceMapping.GetAnnotations().GetFields()[v1alpha1.AnnotationKeyAssetInfo].GetStructValue().AsMap()["iamPolicies"]; !ok {
```
capri-xiyue, line 198: Do we have any iamPolicies set in the gcs bucket?

sailorlqh: Yes, we will be using the gcs bucket where we upload files to do the resource mapping. From https://github.com/abcxyz/pmap/blob/main/terraform/modules/common/main.tf#L167, this bucket is guarenteed to have iamPolicys.

prober/config.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type config struct {
+	ProjectID                string        `env:"PROBER_PROJECT_ID,required"`
+	GCSBucketID              string        `env:"PROBER_BUCKET_ID,required"`
+	BigQueryDataSetID        string        `env:"PROBER_BIGQUERY_DATASET_ID,required"`
+	MappingTableID           string        `env:"PROBER_MAPPING_TABLE_ID,required"`
+	QueryRetryWaitDuration   time.Duration `env:"PROBER_QUERY_RETRY_WAIT_DURATION,default=5s"`
+	QueryRetryLimit          uint64        `env:"PROBER_QUERY_RETRY_COUNT,default=10"`
+	ProberMappingServiceName string        `env:"PROBER_MAPPING_SERVICE_NAME,required"`
```
capri-xiyue, line 32: It make me feel confused what it is, i think here you'd mean the gcs bucket prefix, right? like `mapping` and `policy`?

prober/config.go
```diff
@@ -0,0 +1,42 @@
...
+	ProberMappingFilePrefix  string        `env:"PROBER_MAPPING_FILE_PREFIX,required"`
```
capri-xiyue, line 33: If the previous one is the gcs bucket prefix, what's this one?

prober/main.go
```diff
@@ -0,0 +1,225 @@
...
+	// TODO: add probePolicy() to probe policy service
+	// errors.Join() is used here so we can join the errors from probePolicy and return them together.
+	// if err := probePolicy(ctx, ts); err != nil {
+	// 	probeErr = errors.Join(probeErr, fmt.Errorf("prober failed for policy service: %w", err))
+	// } else {
+	// 	logging.Info("Policy probe failed.")
+	// }
+	if probeErr == nil {
+		logging.Info("Mapping probe successed")
+	}
```
yolocs, line 112: I see now you want to output "mapping success" vs. "policy success". Feel free to use the if/else

yolocs (approved):

prober/main.go
sailorlqh: My current thought only print "Mapping and Policy probe succeed" after probePolicy is added.

sailorlqh:
> Have you manually tried probing to see if it works?

Just, I have tested using current pmap-dev project.

prober/main.go
```diff
@@ -0,0 +1,221 @@
...
+	logging.Info("Mapping probe started")
+
+	traceID := fmt.Sprintf("%s-%s", proberMappingTraceIDPrefix, timestamp)
+	logging.Infof("using traceID: %s", traceID)
+
+	data := []byte(fmt.Sprintf(`
+resource:
+  name: %s/%s
+  provider: %s
+annotations:
```
capri-xiyue, line 128: We need to add a `label: prober` in the annotation to indicates it is for prober so that we can easily filter these entries. I remember we have such similar annotations in the prober design doc.

sailorlqh: Added

capri-xiyue (approved):

PR #161, opened 2023-08-01
feat: update goreleaser for prober

sailorlqh:
The `docker.goreleaser.yaml` is also used by the `ci` workflow, so it will build prober images when `ci` workflow is triggered. However we don't really need a prober in the `ci` project. 

We can create a ci dedicated `ci.docker.goreleaser.yaml` as a follow up if we think the current solution is not tolerable. My personal opinion is the current way is tolerable, having one yaml for both `release` and `ci` workflow to use is better than maintaining two similar yaml files.


prober/prober.dockerfile
```diff
@@ -0,0 +1,7 @@
+# Base image that supports bash
+FROM cgr.dev/chainguard/bash:latest
+
+COPY pmap-prober /pmap-prober
+
+# Run the binary on container startup.
+ENTRYPOINT ["./pmap-prober"]
```
capri-xiyue, line 7: Unlike JVS, you are no longer using simple bash for probder, do you need this?

.goreleaser.docker.yaml
```diff
@@ -87,13 +110,62 @@ dockers:
...
+    goos: 'linux'
+    goarch: 'amd64'
+    image_templates:
+      - '{{ .Env.DOCKER_REPO }}/pmap-prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
+    dockerfile: './prober/prober.dockerfile'
+    build_flag_templates:
+      - '--platform=linux/amd64'
+      - '--pull'
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Privacy data mapping tool'
```
yolocs, line 126: Update description

.goreleaser.docker.yaml
```diff
@@ -87,13 +110,62 @@ dockers:
...
+    goos: 'linux'
+    goarch: 'arm64'
+    image_templates:
+      - '{{ .Env.DOCKER_REPO }}/pmap-prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
+    dockerfile: './prober/prober.dockerfile'
+    build_flag_templates:
+      - '--platform=linux/arm64'
+      - '--pull'
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Privacy data mapping tool'
```
yolocs, line 147: Update description

prober/prober.dockerfile
```diff
@@ -0,0 +1,7 @@
+# Base image that supports bash
+FROM cgr.dev/chainguard/bash:latest
```
yolocs, line 2: We should use distroless image here.

yolocs (changes requested):

prober/prober.dockerfile
```diff
@@ -0,0 +1,7 @@
+# Base image that supports bash
+FROM gcr.io/distroless/base-debian11:latest
```
capri-xiyue, line 2: Why use base-debian, does static work here https://github.com/abcxyz/jvs/blob/main/Dockerfile

yolocs, line 2: Should be `gcr.io/distroless/static`

yolocs (changes requested):

yolocs (approved):
Did you try the prober out manually?

Issue #165, opened 2023-08-03
Create a CI dedicated goreleaser.docker.yaml


sailorlqh:
No need to release prober images in the ci workflow.

capri-xiyue:
@sailorlqh Could you add more context? We still need to be able to validate the prober build process for each ci/cd run.

sailorlqh:
We decided to only release prober with `OS = linux` and `ARCH = amd64` as we don't expect prober to be run in other OS and Arch. We still kept prober release within the CI workflow, because this is an essential part of the CI/CD check.

PR #172, opened 2023-08-04
fix: Update ci workflow to use ci dedicated goreleaser config by default

sailorlqh:
Fix: #165 

Updated CI workflow to use `.goreleaser.ci.docker.yaml` by default to save ci workflow time as releasing for OS other than linux, Arch other then amd64 aren't needed in CI.

Will update the release workflow after this PR is merged to make it run ci workflow with `.goreleaser.docker.yaml` before the it does the github and artifact registry release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,19 @@ on:
     branches:
       - 'main'
   workflow_dispatch:
+    inputs:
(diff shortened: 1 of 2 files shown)


sailorlqh:
Closed this PR because  and opened #173 for a easier fix.

PR #173, opened 2023-08-04
fix: remove some prober release support

sailorlqh:
fix: #172 
discussed offline, decided for prober, we just release amd64 and linux.

### DIFF
--- a/.goreleaser.docker.yaml
+++ b/.goreleaser.docker.yaml
@@ -128,6 +128,9 @@ dockers:
       - '--label=org.opencontainers.image.title=pmap-prober'
       - '--label=org.opencontainers.image.version={{ .Version }}'
 
+archives:
+  - allow_different_binary_count: true
+
 docker_manifests:
   -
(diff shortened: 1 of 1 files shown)


.goreleaser.docker.yaml
```diff
@@ -164,7 +142,6 @@ docker_manifests:
     name_template: '{{ .Env.DOCKER_REPO }}/pmap-prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}'
```
yolocs, line 142: I believe you can remove prober from docker_manifests now since we're not shooting for multiarch image.

.goreleaser.docker.yaml
```diff
@@ -62,12 +62,9 @@ builds:
       - '-X={{ .ModulePath }}/internal/version.Commit={{ .Commit }}'
       - '-extldflags=-static'
     goos:
-      - 'darwin'
       - 'linux'
```
yolocs, line 65: Add a comment say we only expecting running prober from amd64/linux

yolocs (approved):

Issue #164, opened 2023-08-03
Add support for mapping subresource.


sailorlqh:
### TL;DR

Our current approach for doing resource mapping uses [searchAllResource](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/searchAllResources) api provided by [gcloud asset inventory(CAIS)](https://cloud.google.com/asset-inventory/docs/overview). We need to support resource granularity beyond this level e.g. cloud spanner table wise. Subresource is referred as those resources beyond the cloud resource level.

### Detailed design

The proposed way for supporting this is adding a new field called `subscope` in the [Resource proto](https://github.com/abcxyz/pmap/blob/main/protos/v1alpha1/resource_mapping.proto#L45-L52).

#### Subresource representation
The `subscope` field will be a string in the format of `hierarchy_string?qualifier_string`.

The `hierarchy string` will need to reflect the hierarchy of a subresource from larger granularity to narrower granularity with the format of `parent/foo/child/bar`. Example: `instance/ins/dataset/db1`.

The `qualifier_string` is a string that contains the additional information to further help determine the resources as key value pairs: `key1=value1&key2=value2`. Example: `ABC=XYZ&PROVIDER=GCP`.

#### Inheritance and how to query
Subresource will inherit the attributes (e.g. parent info, IAM policy) and annotations from its cloud resource level parent. 

When querying subresource, we consider a match only when resource.name and resource.subscope both match.

sailorlqh:
This is reopened, as we have not yet update the integration tests.

PR #176, opened 2023-08-07
feat: add subscope and validate if subscope is in correct format.

sailorlqh:
fix: #164


README.md
```diff
@@ -1,5 +1,7 @@
 # pmap
 
+
+##temp
```
capri-xiyue, line 4: Is this a typo?

apis/v1alpha1/resource_mapping.pb.go
```diff
@@ -126,6 +126,16 @@ type Resource struct {
 	// Required. The full resource name. Example:
 	// `//spanner.googleapis.com/projects/p1/instances/i1/databases/d1`
 	Name string `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`
+	// Optional. The representation for subresource in the format of a URL.
+	// Subresource is the resource thats beyond the cloud resource level,
```
capri-xiyue, line 130: nits: Subscope is the resource that beyonds cloud resource level.

apis/v1alpha1/resource_mapping.pb.go
```diff
@@ -126,6 +126,16 @@ type Resource struct {
...
+	// like spanner table, GCS object.
+	// Format: "hierarchy_string?qualifier_string"
+	// where hierarchy_string represents the hierarchy of the subresource to this parent
+	// in the format of parent/foo/child/bar.
+	// and qualifier_string represents the additional information of the subresource
+	// in the format of: key1=value1&key2=value2.
+	// Example: rows with provider=gcp in a database table1 would be: `table/table1?provider=gcp`
```
capri-xiyue, line 137: use another example, `provider=gcp` is confusing here as we have `provider` as a field of the `resource_mapping`

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,54 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	if r.Subscope == "" {
+		return nil
+	}
+
+	// normalize subscope string to only have lower cases
+	r.Subscope = strings.ToLower(r.Subscope)
+
+	u, err := url.Parse(r.Subscope)
+	if err != nil {
+		fmt.Println("hahahahahahah")
```
capri-xiyue, line 78: what's this?

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,54 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	return
+}
+
+func validateAndNormalizeSubscope(r *Resource) error {
+	if r.Subscope == "" {
+		return nil
+	}
+
+	// normalize subscope string to only have lower cases
+	r.Subscope = strings.ToLower(r.Subscope)
```
capri-xiyue, line 74: Do we need to normalize it or do we need to do any validation? My concern is that in BQ entry, we will record an reference to the github file and github commit. If we do any changes to the user's input and the bq entry doesn't match the source of truth in github repo, it will bring confusion.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,54 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	u, err := url.Parse(r.Subscope)
+	if err != nil {
+		fmt.Println("hahahahahahah")
+		return fmt.Errorf("failed to parse subscope string: %w", err)
+	}
+	fmt.Println(u)
```
capri-xiyue, line 81: delete it?

apis/v1alpha1/validation_test.go
```diff
@@ -89,6 +91,81 @@ func TestValidateResouceMapping(t *testing.T) {
 				Resource: &Resource{
 					Provider: "gcp",
 					Name:     "//pubsub.googleapis.com/projects/test-project/topics/test-topic",
+					Subscope: "parent/foo/child/bar?key1=value1&key2=value2",
```
capri-xiyue, line 94: add tc.wantSubscope to verify the logic works as expected. Same for other test cases when needed.

capri-xiyue (changes requested):

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,17 @@ message Resource {
   // Required. The full resource name. Example:
   // `//spanner.googleapis.com/projects/p1/instances/i1/databases/d1`
   string name = 2;
+
+  // Optional. The representation for subresource in the format of a URL. 
+  // Subresource is the resource thats beyond the cloud resource level,
```
capri-xiyue, line 54: same

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,17 @@ message Resource {
...
+  // like spanner table, GCS object.
+  // Format: "hierarchy_string?qualifier_string"
+  // where hierarchy_string represents the hierarchy of the subresource to this parent
+  // in the format of parent/foo/child/bar.
+  // and qualifier_string represents the additional information of the subresource
+  // in the format of: key1=value1&key2=value2.
+  // Example: rows with provider=gcp in a database table1 would be: `table/table1?provider=gcp`
```
capri-xiyue, line 61: same

apis/v1alpha1/validation.go
sailorlqh: The normalization here only apply to the subscope string. I added the comment to subscope to let the user know the string will be normalized.

apis/v1alpha1/validation_test.go
sailorlqh: Added.

apis/v1alpha1/resource_mapping.pb.go
sailorlqh: updated to userID=abcxyz

README.md
sailorlqh: Removed.

apis/v1alpha1/validation.go
capri-xiyue: By the way, I think normalize here doesn't make the subscope all lower case when it goes to bigquery. I doubt users won't be catch the normalize part of  proto description when they query bq. Can we change it to validation that if they have non-lower cases, just throw error?

sailorlqh: Updated.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +37,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	return
+}
+
+func validateAndNormalizeSubscope(r *Resource) error {
+	if r.Subscope == "" {
+		return nil
+	}
+
+	// normalize subscope string to only have lower cases
```
capri-xiyue, line 73: This is not normalize, this is just a validation check. In addition, I think we can't enforce value to be lower case, you can just enforce key? for example, gcs object can include upper case.

sailorlqh: Update to only check is keys are in lower case.

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,18 @@ message Resource {
   // Required. The full resource name. Example:
   // `//spanner.googleapis.com/projects/p1/instances/i1/databases/d1`
   string name = 2;
+
+  // Optional. The representation for subresource in the format of a URL.
```
yolocs, line 53: Could you format the comment to limit to 80 characters? Also add a few empty lines in between examples to make it more readable.

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,18 @@ message Resource {
   // Required. The full resource name. Example:
   // `//spanner.googleapis.com/projects/p1/instances/i1/databases/d1`
   string name = 2;
+
+  // Optional. The representation for subresource in the format of a URL. 
+  // Subscope is the resource that beyonds cloud resource level.
+  // like spanner table, GCS object.
+  // Format: "hierarchy_string?qualifier_string"
```
yolocs, line 56: Usually we do
```suggestion
  // Format: "[HIERARCHY]?[QUALIFIERS]"
```

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,18 @@ message Resource {
...
+  // where hierarchy_string represents the hierarchy of the subresource to this parent
+  // in the format of parent/foo/child/bar.
+  // and qualifier_string represents the additional information of the subresource
+  // in the format of: key1=value1&key2=value2.
+  // The subscope string should only have lowercase.
```
yolocs, line 61: "should only have lowercase" - this doesn't match the examples. You obviously have upper cases

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,18 @@ message Resource {
...
+  // Example: rows with userID=abcxyz in a database table1 would be: `table/table1?userID=abcxyz`
+  string subscope = 3;
```
yolocs, line 63: We need to document that the keys should be in order?

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	_, err = url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return fmt.Errorf("failed to parse qualifier string: %w", err)
+	}
+
+	// check if key value pairs are in alphabetical order
+	// and if key is in lower case
+	s := strings.Split(u.RawQuery, "&")
+
+	for i := 0; i < len(s); i++ {
```
capri-xiyue, line 85: This part is difficult to understand. Given a example, subscope is "parent/foo/child/bar?key2=value2&key1=value1", what will be the value of u.RawQuery?

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+func validateResource(r *Resource) (vErr error) {
+	if r.Name == "" {
+		vErr = errors.Join(vErr, fmt.Errorf("empty resource name"))
+	}
+
+	if r.Provider == "" {
+		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
+	}
+
+	if err := validateAndNormalizeSubscope(r); err != nil {
```
yolocs, line 59: You are just validating. There is no normalizing

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	u, err := url.Parse(r.Subscope)
+	if err != nil {
+		return fmt.Errorf("failed to parse subscope string: %w", err)
+	}
+
+	_, err = url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return fmt.Errorf("failed to parse qualifier string: %w", err)
+	}
```
yolocs, line 79: Why parsing the query again? Won't `url.Parse` already cover that?

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	// check if key value pairs are in alphabetical order
+	// and if key is in lower case
+	s := strings.Split(u.RawQuery, "&")
+
+	for i := 0; i < len(s); i++ {
+		s1 := strings.Split(s[i], "=")
+		if s1[0] != strings.ToLower(s1[0]) {
+			return fmt.Errorf("key should only contain lower case letter")
+		}
```
yolocs, line 89: I strongly recommend getting rid of this. This check poses a strong limitation without any known benefit.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +35,64 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	}
+
+	_, err = url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return fmt.Errorf("failed to parse qualifier string: %w", err)
+	}
+
+	// check if key value pairs are in alphabetical order
+	// and if key is in lower case
+	s := strings.Split(u.RawQuery, "&")
```
capri-xiyue, line 83: The naming is really confusion and the validation logic is difficult to understand. I'd suggest you renaming the variable with clear meaning instead of using `s` or `s1`. For example, you can use `hierarchyStr`, qualifierStr`, `qualifierKey`, `qualifierValue`, `hierarchyKey` `hierarchyValue`. In addition, do you need any validation for the `hierarchyKey`?

yolocs, line 83: I propose a different approach.

1. Get the current query via `u.Query()`. Notice it's possible for one key to have multiple values.
2. Generate a "want" query string with keys sorted and values sorted.
3. Compare it to `u.RawQuery` and if they don't match, return error

yolocs (changes requested):

apis/v1alpha1/validation.go
capri-xiyue: @yolocs  By the way,I doubt whether we need to do such validation here like key value pairs are in alphabetical order? I think making keys all lower case could bring some value. But I didn't get the point that  key value pairs are in alphabetical order

sailorlqh: making the key value pairs in alphabetical put can help make sure the qualifier string will be the same as long as the key value pairs are the same, so we won't have cases like `key1=value1&key2=value2` and `key2=value2&key1=value1` both exists.

As for validating hierarchyKey, if we decided not to check key is in lower case, and we can't validate the resource hierarchy, I don't think we need any validation.

apis/v1alpha1/validation.go
sailorlqh: Not really. Query parses RawQuery and returns the corresponding values. It silently discards malformed value pairs. So we need to use ParseQuery to check if there are any errors. see: https://pkg.go.dev/net/url#URL.Query

apis/v1alpha1/validation.go
sailorlqh: Removed

apis/v1alpha1/validation.go
sailorlqh: Done.

protos/v1alpha1/resource_mapping.proto
sailorlqh: Added

apis/v1alpha1/validation.go
```diff
@@ -33,11 +36,71 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	for _, k := range keys {
+		sort.Strings(q[k])
+		for _, v := range q[k] {
+			kvPairs = append(kvPairs, fmt.Sprintf("%s=%s", k, v))
+		}
+	}
+
+	wantQueryString := strings.Join(kvPairs, "&")
+	if wantQueryString != u.RawQuery {
+		return fmt.Errorf("key values pairs should be in alphabetical order, want: %s, got: %s", wantQueryString, u.RawQuery)
```
yolocs, line 102:
```suggestion
		return fmt.Errorf("qualifiers must be in alphabetical order, got: %s, want: %s", u.RawQuery, wantQueryString)
```

apis/v1alpha1/validation_test.go
```diff
@@ -111,6 +193,11 @@ func TestValidateResouceMapping(t *testing.T) {
 			if diff := testutil.DiffErrString(err, tc.expErr); diff != "" {
 				t.Errorf("ValidateResourceMapping got unexpected error: %s", diff)
 			}
+			if tc.wantSubscope != "" {
+				if diff := cmp.Diff(tc.data.Resource.Subscope, tc.wantSubscope); diff != "" {
```
yolocs, line 197: Otherwise you output would be wrong
```suggestion
				if diff := cmp.Diff(tc.wantSubscope, tc.data.Resource.Subscope); diff != "" {
```

apis/v1alpha1/validation.go
yolocs: Not sure I see the value of enforcing lower case keys.

protos/v1alpha1/resource_mapping.proto
```diff
@@ -49,6 +49,20 @@ message Resource {
...
+  // like spanner table, GCS object.
+  // Format: "[HIERARCHY]?[QUALIFIERS]"
+  // where HIERARCHY represents the hierarchy of the subresource
+  // to its parent in the format of parent/foo/child/bar.
+  // and QUALIFIERS represents the additional information of
+  // the subresource in the format of: key1=value1&key2=value2.
+  // the keys and values should be in alphabetical order.
+  // 
+  // Example: rows with userid=abcxyz in a database table1 would be:
+  // `table/table1?userid=abcxyz`
```
yolocs, line 64:
```suggestion
  // Optional. The representation for subresource in the format of a URL. 
  // Subscope is the resource that is beyond cloud resource level, e.g.
  // spanner table, GCS object, etc.
  //
  // Format: "[SUBRESOURCE_HIERARCHY]?[ADDITIONAL_QUALIFIERS]"
  //
  // [SUBRESOURCE_HIERARCHY] represents the hierarchy of the subresource
  // E.g. "databases/example-db/tables/example-table".
  //
  // [ADDITIONAL_QUALIFIERS] represents the additional qualifiers to select
  // the data in the subresource. E.g. "source=example-org1&team=example-team".
  // Notice the keys and values must be in alphabetical order.
  // 
  // Full example: "databases/example-db/tables/example-table?source=example-org1&team=example-team".
```

yolocs (changes requested):

apis/v1alpha1/validation.go
capri-xiyue: Discussed offline, I think we only need to enforce the key value pairs are in alphabetical order.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +36,71 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
-	if m.Resource.Provider == "" {
-		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
-	}
+
 	if _, ok := m.Annotations.AsMap()[AnnotationKeyAssetInfo]; ok {
 		vErr = errors.Join(vErr, fmt.Errorf("reserved key is included: %s", AnnotationKeyAssetInfo))
 	}
+
+	if err := validateResource(m.Resource); err != nil {
+		vErr = errors.Join(vErr, fmt.Errorf("invalid Resource: %w", err))
```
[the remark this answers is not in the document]
capri-xiyue: Please double check the error returned in validateResource is clear enough.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +36,71 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+	if r.Name == "" {
+		vErr = errors.Join(vErr, fmt.Errorf("empty resource name"))
+	}
+
+	if r.Provider == "" {
+		vErr = errors.Join(vErr, fmt.Errorf("empty resource provider"))
+	}
+
+	if err := validateSubscope(r); err != nil {
+		vErr = errors.Join(vErr, fmt.Errorf("invalid subscope: %w", err))
```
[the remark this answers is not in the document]
capri-xiyue: Please double check the error returned in validateSubscope is clear enough. I feel  `fmt.Errorf("failed to parse qualifier string: %w", err)` and `fmt.Errorf("key values pairs should be in alphabetical order, want: %s, got: %s", wantQueryString, u.RawQuery)` is not clear enough when you don't wrap error here.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +36,71 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	keys := make([]string, 0, len(q))
+	for k := range q {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var kvPairs []string
+	for _, k := range keys {
+		sort.Strings(q[k])
```
capri-xiyue, line 94: Why do you need to sort here? I thought the key should be in order not the values

capri-xiyue (changes requested):

apis/v1alpha1/validation.go
sailorlqh: Yes, otherwise if a key has multiple values, it would cause problem. For example: `key1=value1&key1=value2&key2=value3` and `key1=value2&key1=value1&key2=value3`

apis/v1alpha1/validation.go
sailorlqh: No longer check if keys are in lower case.

apis/v1alpha1/validation.go
```diff
@@ -33,11 +36,71 @@ func ValidateResourceMapping(m *ResourceMapping) (vErr error) {
...
+
+	// If r.Subscope = "parent/foo/child/bar?key1=value1&key2=value2"
+	// after url.Parse(r.Subscope), we will have:
+	// u.RawQuery: key1=value1&key2=value2
+	u, err := url.Parse(r.Subscope)
+	if err != nil {
+		return fmt.Errorf("failed to parse subscope string: %w", err)
+	}
+
+	// ParseQuery is needed here to check if query string has errors
```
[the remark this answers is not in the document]
sailorlqh: Done.

yolocs (approved):

capri-xiyue (approved):

PR #184, opened 2023-08-10
fix: add integration test case for subresource mapping

sailorlqh:
fixes: #164

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -110,13 +110,13 @@ func TestMain(m *testing.M) {
 
 // TestMappingEventHandling tests the entire flow from uploading a mapping data
 // to GCS, triggering pmap event handler, to writing it to a BigQuery table.
-// TODO(#56): add validation logic for github source attributes as object metadata.
 func TestMappingEventHandling(t *testing.T) {
 	t.Parallel()
 	cases := []struct {
 		name                 string
 		resourceName         string
 		bigqueryTable        string
+		subscope             string
 		wantCAISProcessed    bool
 		wantGithubSource     *v1alpha1.GitHubSource
 		wantResourceMapping  *v1alpha1.ResourceMapping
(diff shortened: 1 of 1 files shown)


test/integration/main_test.go
```diff
@@ -188,6 +188,30 @@ func TestMappingEventHandling(t *testing.T) {
 			},
 			wantProcessErrSubStr: "failed to validate and enrich",
 		},
+		{
```
capri-xiyue, line 191: I think you can just use resource mapping with subscope as the test case. No need to have one resource mapping without subscope and the other resource mapping with subscope.

test/integration/main_test.go
```diff
@@ -217,6 +242,19 @@ contacts:
   email:
   - group@example.com
 `, tc.resourceName, traceID.String()))
+			} else {
```
capri-xiyue, line 245: This if else is a little bit ugly. Usually you need to do refactor to avoid it.

test/integration/main_test.go
sailorlqh: Updated.

test/integration/main_test.go
sailorlqh: This is removed since now we add subscope to existing test cases.

capri-xiyue (changes requested):

test/integration/main_test.go
capri-xiyue: You don't want to add subscope for all test cases including failure cases, right?

test/integration/main_test.go
```diff
@@ -206,17 +210,17 @@ func TestMappingEventHandling(t *testing.T) {
...
   name: %s
   provider: gcp
+  subscope: %s
 annotations:
   traceID: %s
 contacts:
   email:
   - group@example.com
-`, tc.resourceName, traceID.String()))
+`, tc.resourceName, tc.subscope, traceID.String()))
```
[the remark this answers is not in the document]
capri-xiyue: I took a look at the code. Probably the refactor takes time.  I think currently we can enforce all test cases to have subscope and add a TODO for it.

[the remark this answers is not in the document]
sailorlqh: Created #185 for follow up.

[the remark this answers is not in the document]
capri-xiyue: Can you add a comment which mentions the TODO? 
`// TODO(#185): Refactor so we don't have to enforce subscope in every test case.`

test/integration/main_test.go
```diff
@@ -110,13 +110,13 @@ func TestMain(m *testing.M) {
 
 // TestMappingEventHandling tests the entire flow from uploading a mapping data
 // to GCS, triggering pmap event handler, to writing it to a BigQuery table.
-// TODO(#56): add validation logic for github source attributes as object metadata.
```
capri-xiyue, line 113: Add a comment here 
// TODO(#185): Refactor so we don't have to enforce subscope in every test case.

sailorlqh: Added

capri-xiyue (approved):

PR #174, opened 2023-08-07
feat: added monitoring policy

capri-xiyue:
fixed #156

PR #186, opened 2023-08-10
feat: add alerting to e2e module

sailorlqh:
fix: #174

### DIFF
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -19,7 +19,8 @@ locals {
   alert_subscription_map = {
     mapping_bigquery         = module.common_infra.bigquery_subscriptions[local.mapping_service_name].event_subscription_id,
     mapping_bigquery_failure = module.common_infra.bigquery_subscriptions[local.mapping_service_name].failure_subscription_id,
-    policy_bigquery          = module.common_infra.bigquery_subscriptions[local.policy_service_name].failure_subscription_id,
+    policy_bigquery          = module.common_infra.bigquery_subscriptions[local.policy_service_name].event_subscription_id,
+    policy_bigquery_failure  = module.common_infra.bigquery_subscriptions[local.policy_service_name].failure_subscription_id,
     mapping_gcs              = module.mapping_service.gcs_notification_subscription_id,
     policy_gcs               = module.policy_service.gcs_notification_subscription_id
   }
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -16,13 +16,15 @@ locals {
   mapping_service_name = "mapping"
   policy_service_name  = "policy"
 
-  alert_subscription_map = {
-    mapping_bigquery         = module.common_infra.bigquery_subscriptions[local.mapping_service_name].event_subscription_id,
-    mapping_bigquery_failure = module.common_infra.bigquery_subscriptions[local.mapping_service_name].failure_subscription_id,
-    policy_bigquery          = module.common_infra.bigquery_subscriptions[local.policy_service_name].failure_subscription_id,
-    mapping_gcs              = module.mapping_service.gcs_notification_subscription_id,
-    policy_gcs               = module.policy_service.gcs_notification_subscription_id
-  }
+  alert_subscription_list = [
+    module.common_infra.bigquery_subscriptions[local.mapping_service_name].event_subscription_id,
+    module.common_infra.bigquery_subscriptions[local.mapping_service_name].failure_subscription_id,
+    module.common_infra.bigquery_subscriptions[local.policy_service_name].event_subscription_id,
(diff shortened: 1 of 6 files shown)


terraform/modules/monitoring/alert.tf
```diff
@@ -112,11 +112,11 @@ resource "google_monitoring_alert_policy" "pmap_messages_too_old" {
 }
 
 resource "google_monitoring_alert_policy" "pmap_num_of_undeliverable_messages_oversized" {
-  for_each = toset(var.pmap_subscription_ids)
+  for_each = var.pmap_subscription_ids
 
   project = var.project_id
 
-  display_name = "Nnumber of undeliverable messages for pmap subscription id {${each.key}} oversized[MEAN]"
+  display_name = "Nnumber of undeliverable messages for pmap subscription id {${each.value}} oversized[MEAN]"
```
yolocs, line 119:
```suggestion
  display_name = "Number of undeliverable messages for pmap subscription id {${each.value}} oversized[MEAN]"
```

terraform/e2e/main.tf
```diff
@@ -15,6 +15,14 @@
 locals {
   mapping_service_name = "mapping"
   policy_service_name  = "policy"
+
+  alert_subscription_map = {
+    mapping_bigquery         = module.common_infra.bigquery_subscriptions[local.mapping_service_name].event_subscription_id,
+    mapping_bigquery_failure = module.common_infra.bigquery_subscriptions[local.mapping_service_name].failure_subscription_id,
+    policy_bigquery          = module.common_infra.bigquery_subscriptions[local.policy_service_name].failure_subscription_id,
```
capri-xiyue, line 22: Should we have both policy_bigquery and policy_bigquery_failure?

terraform/modules/monitoring/alert.tf
```diff
@@ -78,11 +78,11 @@ resource "google_monitoring_alert_policy" "prober_service_success_number_below_t
 }
 
 resource "google_monitoring_alert_policy" "pmap_messages_too_old" {
-  for_each = toset(var.pmap_subscription_ids)
+  for_each = var.pmap_subscription_ids
```
capri-xiyue, line 81: Why change this?

terraform/modules/monitoring/variables.tf
```diff
@@ -102,8 +102,8 @@ variable "log_level" {
 }
 
 variable "pmap_subscription_ids" {
-  type        = list(string)
-  default     = []
+  type        = map(string)
```
capri-xiyue, line 105: I think a list will be more user friendly.

terraform/e2e/main.tf
```diff
@@ -15,6 +15,14 @@
 locals {
   mapping_service_name = "mapping"
   policy_service_name  = "policy"
+
+  alert_subscription_map = {
```
capri-xiyue, line 19: Why use map here?

capri-xiyue, line 19: If you want to introduce more readability in the e2e module, it is ok that you can local map here. But you should not enforce all users to have a map when using monitoring module.

terraform/modules/monitoring/variables.tf
sailorlqh: Terraform will complain and throw out errors if it’s a list and says it prefers a map. I have tried using a list, it didn’t work, so I changed to this.

sailorlqh: If map isn’t wanted, I can try if set works.

capri-xiyue: I ran terraform apply and it worked without any warning.

capri-xiyue: Discussed offline, the original module should work as I changed the list to set in the monitoring module.

sailorlqh: Updated.

terraform/e2e/main.tf
sailorlqh: Added

terraform/e2e/main.tf
sailorlqh: changed to list.

yolocs (approved):
lgtm. Will let Xiyue double check

terraform/modules/monitoring/variables.tf
sailorlqh: Discussed offline, reproduced the real error, and decided to use map and some comments to describe why we used map here.

terraform/e2e/main.tf
capri-xiyue: Did you add it?

capri-xiyue (approved):

terraform/e2e/main.tf
sailorlqh: sorry it's somehow override by other changes, added now.

PR #187, opened 2023-08-11
chore: add some comments

sailorlqh:
I think this would be helpful for users if they set the scope to folder level but didn't grant the service account the asset viewer role to the folder level.

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -47,8 +47,9 @@ type HandlerConfig struct {
 type MappingHandlerConfig struct {
 	// DefaultResourceScope is the default resource scope to search resources.
 	// This is only used for global resources such as GCS bucket.
-	// Please make sure your run-pmap service account was granted the permission
-	// as a Cloud Asset Viewer role to corresponding scope level.
+	// Please make sure the Service Account used in the Cloud Run service for
+	// Data Mapping is granted the roles/cloudasset.viewer to the corresponding
+	// scope level.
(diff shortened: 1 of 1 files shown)


pkg/server/config.go
```diff
@@ -47,6 +47,8 @@ type HandlerConfig struct {
 type MappingHandlerConfig struct {
 	// DefaultResourceScope is the default resource scope to search resources.
 	// This is only used for global resources such as GCS bucket.
+	// Please make sure your run-pmap service account was granted the permission
```
capri-xiyue, line 50: Nits: Please make sure the Service Account used in the Cloud Run service for Data Mapping is granted the `roles/cloudasset.viewer` to the corresponding scope level.

 run-pmap service account is confusing for people who are not familiar with the implementations details of  our terraform module

sailorlqh: Thanks! Updated.

pkg/server/config.go
```diff
@@ -47,6 +47,9 @@ type HandlerConfig struct {
 type MappingHandlerConfig struct {
 	// DefaultResourceScope is the default resource scope to search resources.
 	// This is only used for global resources such as GCS bucket.
+	// Please make sure the Service Account used in the Cloud Run service for
+	// Data Mapping is granted the roles/cloudasset.viewer to the corresponding
```
capri-xiyue, line 51: make `roles/cloudasset.viewer` clear. Probably a single quote

sailorlqh: Done

capri-xiyue (approved):

PR #199, opened 2023-09-11
fix: added missing dependency

capri-xiyue:
added missing dependency in prober

### DIFF
--- a/terraform/modules/monitoring/prober.tf
+++ b/terraform/modules/monitoring/prober.tf
@@ -26,6 +26,29 @@ locals {
     "LOG_LEVEL" : var.log_level
   }
 }
+
+resource "google_project_service" "serviceusage" {
+  project = var.project_id
+
+  service            = "serviceusage.googleapis.com"
+  disable_on_destroy = false
+}
+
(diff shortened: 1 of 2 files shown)


sqin2019 (dismissed):

terraform/modules/monitoring/prober.tf
```diff
@@ -34,6 +57,10 @@ resource "google_cloud_run_v2_job" "pmap_prober" {
 
   location = "us-central1"
 
+  depends_on = [
+    google_project_service.prober_services["cloudscheduler.googleapis.com"],
```
sqin2019, line 61: actually, it is better to add this dependency in resource "google_cloud_scheduler_job" "job"?

capri-xiyue: cloud scheduler job depends on `google_cloud_run_v2_job.pmap_prober`

capri-xiyue: Thanks for catching it. This is a typo.

sqin2019 (approved):

PR #202, opened 2023-09-13
fix: added prober output in e2e

capri-xiyue:
added prober output in e2e


terraform/e2e/outputs.tf
```diff
@@ -42,6 +42,16 @@ output "run_service_account_name" {
   value       = module.common_infra.run_service_account_name
 }
 
+output "prober_service_account" {
```
sqin2019, line 45: Should we add service account member just in case?

sqin2019 (approved):

PR #208, opened 2023-11-02
chors: use t.Cleanup instead of defer for cleaning up

sailorlqh:
I think this is the preferred way.

### DIFF
--- a/pkg/server/event_handler_test.go
+++ b/pkg/server/event_handler_test.go
@@ -46,8 +46,7 @@ func TestEventHandler_NewHandler(t *testing.T) {
 	ctx := context.Background()
 
 	// Setup fake storage client.
-	hc, done := newTestServer(testHandleObjectRead(t, []byte("test")))
-	defer done()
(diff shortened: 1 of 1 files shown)


pkg/server/event_handler_test.go
```diff
@@ -537,7 +535,8 @@ isOK: true`),
 }
 
 // Creates a fake http client.
-func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.Client, func()) {
+func newTestServer(t *testing.T, handler func(w http.ResponseWriter, r *http.Request)) *http.Client {
```
sethvargo, line 538: Nit: prefer accepting `testing.TB` (interface) instead of `*testing.T` (type). The interface makes it reusable for benchmarks.

sailorlqh: Done, thanks!

sethvargo (approved):

capri-xiyue (approved):

PR #218, opened 2024-03-14
fix: fixed lint

capri-xiyue:


pkg/server/event_handler.go
```diff
@@ -230,6 +230,7 @@ func (h *EventHandler[T, P]) HTTPHandler() http.Handler {
 			Data:       m.Message.Data, // Notification payload.
 			Attributes: m.Message.Attributes,
 		}
+		//nolint:sloglint
```
sqin2019, line 233: Why this is needed?

capri-xiyue: Linters complain. it doesn't allow to use snakecase. But `bucketID` is snakecase

sqin2019 (approved):

PR #230, opened 2025-02-14
Update to Go 1.24 and all deps

sethvargo:


verbanicm (dismissed):
@sethvargo  somethings up with the status checks

